PR #351, opened 2021-03-29
Adding additional point to display the orderbook until the end

### LINKED ISSUE #283: Chart not readable
The chart is not showing how it should be. 

The volume of the bids should extend until the y axis (x=0) and have a green shade in all the area. 

Example of current wrong version in [auction 58](https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=58&chainId=4#topAnchor):

Screenshot 2021-03-24 at 00 16 36

The objective is that it looks something like this (but with current price centered):
Screenshot 2021-03-24 at 00 19 54

As an example, you can try to picture a normal orderbook, but that instead of having a space between the orders (called spread), t**he orders must overlap**. This is a picture of a normal orderbook from Binance: 
Screenshot 2021-03-24 at 00 25 08

**Examples of auctions with unreadable chart**

- As another example, [this auction also](https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=43&chainId=4#topAnchor) presents odd orderbook behaviour, and should attempt to look at explained above. 

- This [auction](https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=1&chainId=4#topAnchor) also shows no orders in the book, or current price.

josojo:
closes #283

I tried out different values and actually, I like it better if the line does not go until the end of the chart.

Here is one of the auction that was previously not nicely displayed: 
image

If we wanna display the green line until the end, all we need to change is: 
```
  // Reduce the min in a 5%
  priceAxis.min = min - min * 0.05
```

### DIFF
--- a/src/components/auction/OrderbookWidget/index.tsx
+++ b/src/components/auction/OrderbookWidget/index.tsx
@@ -83,18 +83,22 @@ const processData = (
     pricePoints.sort((lhs, rhs) => -1 * (lhs.price - rhs.price))
 
     pricePoints.push({
-      price: (highestValue * 101) / 100,
+      price: (highestValue * 104) / 100,
       volume: 0,
     })
 
     pricePoints.sort((lhs, rhs) => -1 * (lhs.price - rhs.price))
+    pricePoints.push({
+      price: (lowestValue * 96) / 100,
+      volume: 0,
+    })
   } else {
     pricePoints.push({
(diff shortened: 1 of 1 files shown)


josojo:
Very valid critiques. 

I will merge this PR anyways, as it is already an improvement on one side of the chart :) 

Your concerns are documented here: 
https://github.com/gnosis/ido-ux/issues/384

PR #387, opened 2021-03-31
(Feature) Responsive views

### LINKED ISSUE #55: New Design

### LINKED ISSUE #75: [UI] Responsive views
I created this issue so I don't forget to check how the app looks on mobile (and to add some new designs when needed).

### LINKED ISSUE #380: Browser Icon Wrong Name
In the tab window, the name has no space. 

Screenshot 2021-03-30 at 12 17 26

It should be Gnosis Auction, with a space between the words.

gabitoesmiapodo:
Closes #55 
Closes #75
Closes #380 

OK, this is not perfect but it'll work (I think).

Tested it on a desktop browser's responsive view emulation, and also using Opera Mobile (which features a crypto wallet) on my cellphone.

I could use the app, and it looks OK.

I couldn't use the Confirm button when trying to place an order, though. The confirmation window just closes... and nothing happens. Maybe it's a browser / wallet problem? I had no other issues but this one.

Also, I'm not sure about wallet connecting / disconnecting on mobile. From what I understand you can't "connect" to the wallet on the browser, you just accept to use it (so disconnecting would make no sense, but I don't know).

Anyways: it "works", and we can probably merge this and fix any issues later.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -128,57 +128,6 @@
         }
       }
 
-      #temporaryMobileBlocker {
-        align-items: center;
-        background-color: #001429;
-        display: flex;
-        flex-direction: column;
-        font-family: Averta, Arial, 'Helvetica Neue', 'Segoe UI', Oxygen, Ubuntu, Cantarell, 'Fira Sans', 'Droid Sans', -apple-system, BlinkMacSystemFont, sans-serif;
-        height: 100vh;
-        justify-content: center;
-        left: 0;
-        padding: 25px 20px;
-        position: fixed;
-        top: 0;
-        width: 100vw;
-        z-index: 12345;
-      }
-
-      #temporaryMobileBlocker .lockIcon {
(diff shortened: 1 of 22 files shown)


gabitoesmiapodo:
Additionally I tested this on Metamask's app for Android. Everything went smoothly:

- When you open the URL, it connects you automatically.
- If you disconnect the wallet, you can connect again following the usual procedure.
- Wallect connect's modal seems to work, but I don't think you are supposed to use it here...
- I could place several orders, cancel orders, etc.
- No noticeable performance issues (mine is a old-ish 4 years old cellphone).

So: it works there. We should test on other apps (Brave, Opera Mobile, etc.) and fix what we find, but it's nice to know that it works.

josojo:
Cool. Looks good to me. It worked also for me in Brave with the a small screen on my desktop :)

josojo (approved):

PR #388, opened 2021-03-31
Make search of pending more specific

### LINKED ISSUE #369: Approving button is not updated
After the approval was mined, the approval was not updated and I could nto proceed with placing an order on mainnet: 
image

hpmaxi:
Closes #369 

This is an idea taken from the original repository, seems like when a stuck tx is found `useHasPendingApproval` returns true (this also returns false when the amount is not specified), I'm not completely sure that this was the problem in mainnet but being more common to have stuck tx in mainnet I would give a try to this approach.

### DIFF
--- a/src/state/transactions/hooks.tsx
+++ b/src/state/transactions/hooks.tsx
@@ -72,16 +72,31 @@ export function useAllTransactions(): { [txHash: string]: TransactionDetails } {
   return transactionsFilteredByConnectedAccount
 }
 
+/**
+ * Returns whether a transaction happened in the last day (86400 seconds * 1000 milliseconds / second)
+ * @param tx to check for recency
+ */
+export function isTransactionRecent(tx: TransactionDetails): boolean {
+  return new Date().getTime() - tx.addedTime < 86_400_000
+}
+
 // returns whether a token has a pending approval transaction
 export function useHasPendingApproval(tokenAddress?: string): boolean {
   const allTransactions = useAllTransactions()
-  return typeof tokenAddress !== 'string'
-    ? false
-    : Object.keys(allTransactions).some((hash) => {
-        if (allTransactions[hash]?.receipt) {
+  return useMemo(() => {
+    return (
+      typeof tokenAddress === 'string' &&
+      Object.keys(allTransactions).some((hash) => {
+        const tx = allTransactions[hash]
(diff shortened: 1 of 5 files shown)


mariano-aguero (approved):

src/state/transactions/hooks.tsx
```diff
@@ -72,16 +72,31 @@ export function useAllTransactions(): { [txHash: string]: TransactionDetails } {
...
+      Object.keys(allTransactions).some((hash) => {
+        const tx = allTransactions[hash]
+        if (!tx) return false
+        if (tx.receipt) {
           return false
         } else {
-          return allTransactions[hash]?.approvalOfToken === tokenAddress
+          const { approvalOfToken } = tx
+          if (!approvalOfToken) return false
+          return approvalOfToken === tokenAddress && isTransactionRecent(tx)
```
josojo, line 97: Can you explain this? You are adding another constraint , so your funciton is returning less likely a true.

But we have the problem that it is not returning true, although it should return true. Hence, I think that your fix will not improve the situation.

hpmaxi: We should make that a user has a pending approval less likely to be true. Maybe I'm missing something, but I will re-check everything.

josojo: Ah I see.

josojo: Also the spender could be the problem: 
https://github.com/Uniswap/uniswap-interface/blob/8fd894f2d1b03921c5519c050aed164343c47fb1/src/state/transactions/hooks.tsx#L80

Maybe we should use uniswaps new logic for it...

josojo:
FYI: 
we also have the current hack in the code:
```
// eslint-disable-next-line no-warning-comments
  // Todo: research why this hack of allowance reload for each block
  // is needed to update the allowance after approving.
  const allowance = useSingleCallResult(contract, 'allowance', inputs, {
    blocksPerFetch: 1,
  }).result
  ```
  
 Uniswap just uses:
 ```
 
  const allowance = useSingleCallResult(contract, 'allowance', inputs).result
  
  
```

Maybe its also related to that....

josojo (approved):
I wasn't able to reproduce tthis bug on mainnet. I think we can merge it like that.

PR #392, opened 2021-04-01
Display locked symbol only when needed

josojo:
The current dev branch has only an ugly solution, which I implemented during the boson call. I think this is the right way to display the auction symbol only at the right time.


josojo:
Merging already, such that customer can test. I tested it several times on rinkeby

PR #396, opened 2021-04-01
Update documentation

mariano-aguero:
No issue related


josojo:
Thanks @mariano-aguero 

But let's not yet merge it. I think we should update all the screenshots and we should also not tell our customers to use a staging verison.

I think we will publish a new version via ipfs, fix the documentation and then we can go live with the documentaiton

josojo:
Code was merged into: https://github.com/gnosis/ido-ux/pull/518

PR #408, opened 2021-04-03
Displaying correct min buy

### LINKED ISSUE #403: Min received amount is incorrect
Seen value: 15B
Expected value: It should be 30/2010

josojo:
closes #403


elena-zh:
The changed LGTM!

PR #409, opened 2021-04-03
tiny fix in price calcuation function

### LINKED ISSUE #404: Chart shows nothing when I place an order with a greater price
1. Here the case: the last price value in the chart = 10.5
2. I place an order with price = 11

**AR**: nothing is displayed in the chart. 
https://drive.google.com/file/d/1hazbv9WsogZWxPl-VECPN522Gp2p_RrS/view
Moreover, chart is not displayed after adding this order

**ER**:  orange area should be displayed before adding an order;
chart is displayed after adding the order

josojo:
closes #404 

The issue was that the clearing price displayed a value of 0. Something that should never happen. Fixed price calculation.


elena-zh:
The changes LGTM!

PR #410, opened 2021-04-03
fixing chart for boson auction

josojo:
When our customerr was running the auction, they would get:
https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=14&chainId=4#topAnchor
image
and it looked like only 2k volume was placed. But indeed 20k was placed

Now its fixed

### DIFF
--- a/src/components/auction/Charts/XYChart.ts
+++ b/src/components/auction/Charts/XYChart.ts
@@ -52,7 +52,7 @@ export const XYChart = (props: XYChartProps): am4charts.XYChart => {
   const numberFormatter = new am4core.NumberFormatter()
   numberFormatter.numberFormat = '#.00a'
   numberFormatter.bigNumberPrefixes = [
-    { number: 1e4, suffix: 'K' }, // Use K only with value greater than 9999.00
+    { number: 1e3, suffix: 'K' }, // Use K only with value greater than 999.00
     { number: 1e6, suffix: 'M' }, // Million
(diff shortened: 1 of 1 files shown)


jarcodallo:
@josojo, @Rafanator asked for this. Could talk with him?

Rafanator:
@jarcodallo Which part do you refer to?

jarcodallo:
> @jarcodallo Which part do you refer to? 

The change that @josojo is proposing will display 1000 as 1k. Currently only if the number is greater than 9999 shows a K suffix. For example:

Currently:
2345.00 -> same value.
123456 -> 12.34k

After this change:
2345.00 -> 2.34k
123456 -> 12.34k

Rafanator:
Thanks! I think we should still leave it as we discussed: 

x > 9,999 --> we use "k"
z <= 9,999 --> we leave as is

josojo:
The correct description would be:

Currently:
2345.00 -> same value.
123456 -> 12.34k

After this change:
2345.00 -> 2.34k
123456 -> 123.4k


>123456 -> 12.34k
Is just plain wrong.

>Thanks! I think we should still leave it as we discussed:
x > 9,999 --> we use "k"
z <= 9,999 --> we leave as is

If it would be easy to implement, I would do it. But I am finding here just a practical way to get rid of the errors that are bugging our customers

mariano-aguero (approved):

PR #411, opened 2021-04-03
Eliminates infura calls, if no account is available

### LINKED ISSUE #401: Unconnected an a ton of infura requests
We really need to fix tthis.
If 10k users connect and everyone creates all these requests, then infura will stop serving us.

josojo:
I think it is fine, we keep infura asking for new blocks, once a user is connected FOR NOW. 
Later on, we can refactor it cleaner

closes #401

Basically, this PR prevents the block number updates and all related actions, as long as no users is connected

### DIFF
--- a/src/state/application/updater.ts
+++ b/src/state/application/updater.ts
@@ -11,7 +11,7 @@ import { updateBlockNumber } from './actions'
 const logger = getLogger('Updater')
 
 export default function Updater() {
-  const { chainId, library } = useActiveWeb3React()
+  const { account, chainId, library } = useActiveWeb3React()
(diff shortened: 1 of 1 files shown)


elena-zh:
Hi @josojo , 
Great job! 
Now I do not see too many 403 errors in console. 
However, I am able to see 4 of them that appear each time when I reload a page. 
They start to appear when I connect to a wallet, then disconnect. Is this OK for now?


Thanks!

josojo:
I am merging this one already, such that we spot potentially introduce errors quicker. I created another issue for elena's findingss

PR #413, opened 2021-04-05
Create devguide02.md

avsavsavs:


src/docs/files/devguide02.md
```diff
@@ -2,36 +2,16 @@
...
 
-8. Bidders who specified a maximum price in their limit order that is equal to or greater than the final clearing price receive tokens at the clearing price. 
-
-9. Bidders that specified a maximum price in their limit order that is less than the final clearing price do not receive the tokens being auctioned, but they can still withdraw their bidding tokens from the DEX. 
-
-Let's take a deeper look at the example provided in the diagram below.
-
-
-The following diagram shows an auction in which 14 tokens are offered at 2 blocks per share:
-
```
josojo, line 25: do you wanna delete all of that without substitution?

avsavsavs: yes, this is part of the announcement post that I added without checking in with @Rafanator. We agreed that it would be better to keep the section as simple as possible as it was created originally. Maybe, if possible, its best if we wait for tomorrow to approve this pull request so @Rafanator can weigh in.

josojo: I think we should at least that something like that:

8. Every user that had a limit price higher than the clearing price can claim the auctioned tokens priced at the clearing price. All other users with orders lower that the clearing price will claim their initial investments back...

avsavsavs: Although this was not in the original section, I think it can help clarify an important part without needing to go to the FAQ section.

josojo: What is the current sentiment? I felt like I needed to explain it too so many people who really wanted to know what its going on that I defintively votte for the longer version

Can we close this. PR/

Rafanator: If you both think it is better, let's do it.

josojo:
Closing as we want a more desciptive readme according to my udnerstanding

PR #420, opened 2021-04-05
(Fix) Remove cancelation date when there are no orders

### LINKED ISSUE #394: Do not show an order cancellation info above 'Your Orders' section when there are no orders
It seems to me a bit illogical to show an order cancellation  info above 'Your Orders' section when there are no orders: 

It would be nice not to show it until an order appears in the section

gabitoesmiapodo:
Closes #394


josojo (approved):
LGTM

PR #435, opened 2021-04-05
[issue-379] xdai gas price is not always set correctly

jarcodallo:
image

image


josojo:
Will thiss PR change the behaviour on mainnet? What is the gas price on mainnet?

It would be great to have also a solution for approval and claiming txs

jarcodallo:
@josojo ready for review

mariano-aguero (approved):

josojo:
Reverting just to be super cautious and to not risk disturbing the sale. Actually, the code is looking good!

elena-zh:
Re-opened as changes were reverted in #472 pr.

PR #436, opened 2021-04-06
displaying quote and base currency correctly in eventdetails

josojo:


josojo:
Rafa tested it and said that it is okay

PR #441, opened 2021-04-06
Log extra data for stuck approval

hpmaxi:
All tx from the user are logged and also the ones that can be returning that an approval tx is pending, more than one could be an error (I'm not taking into account the timestamp in the log just to have more data available)

An extra hypothesis is that the tx, in some cases, doesn't have the `receipt` updated after being mined. This could be checking having the hash of the tx mined as seen in metamask and the last logged one.

A look into `Updater` function and seems solid, more details about the user flow before getting into this error would be very valuable as @mariano-aguero said.

The only thing left to say is that when I had this error in rinkeby, and had some extra logging, these logs appear a lot of times (thousands in a few minutes) like something running more than once (When it works this doesn't happen).


josojo (approved):
Thanks. LGTM

PR #444, opened 2021-04-07
(Fix) Responsive issues

### LINKED ISSUE #415: Mobile view: UI issues
1. Auction details may look ugly when screen width <400 px: 

2. **The issue is reproducible only on  real devices** (iPhone 12 mini and Huawei p20): auction details exceed grid width --> impossible to tap on a tooltip

3. Chart is too small on iPhone 12 mini

4. It would be nice to add a 'Back' button at the top to navigate from an auction details to the auctions page (currently, a user is able to go back by clicking on a device's back button (android) and swipe 9IOS) only. 

5. Timing on a featured widget looks ugly when screen width < 303 px

6. Square Icon on the Home page has dotty border (actually, it does not related to the mobile view...)

### LINKED ISSUE #419: UI for tablets
ON my tablet, the UI starting page looks like that:

Could we please try to display the 'time until auction end' somehow nicer? Maybe it is the easiest solution to display only 2 featured auctions or 1 per row.

gabitoesmiapodo:
Closes #415 
Closes #419

There's still room for improvement, but this should look better than before.

- #419 should be fixed.
- _Auction details may look ugly (...)_ -> Info should look better now.
- _Auction details exceed grid width (...)_ -> Should be fixed.
-  _Impossible to tap on a tooltip (...)_ -> There's not much we can do about this, we probably need to think some other solution for mobile.
- _Chart is too small on iPhone 12 mini (...)_ -> Not really a lot we can do here (especially without an iPhone 12 mini available). It looks fine in the several devices I used, though.
- _It would be nice to add a 'Back' button (...)_ -> No, that would be a nightmare. They can use the browser's back button, or the navigation menu.
- _Timing on a featured widget looks ugly when screen width < 303 px_ -> This should be fixed for most screen resolutions. We don't aim for anything below 320px wide, though (that's iPhone 4 / 5 wide). Anything less would be pretty unusual.
- _Square Icon on the Home page has dotty border (...)_ -> There's nothing to fix, that's how it looks.

### DIFF
--- a/src/components/navigation/sections.tsx
+++ b/src/components/navigation/sections.tsx
@@ -1,12 +1,7 @@
-import React from 'react'
-
-import { SendIcon } from '../icons/SendIcon'
-
 export const navItems = [
   {
     title: 'Auctions',
     url: '/overview#topAnchor',
-    icon: <SendIcon />,
   },
   {
     title: 'Docs',
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -135,7 +135,7 @@ const TimerWrapper = styled.div`
 const TokenSymbol = styled.span`
   align-items: center;
   display: flex;
-  font-size: 16px;
+  font-size: 15px;
   justify-content: center;
 
   & > * {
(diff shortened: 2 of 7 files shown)


josojo (approved):
You rock!

I tested it and it looks so much better

src/components/navigation/sections.tsx
```diff
@@ -1,12 +1,7 @@
-import React from 'react'
-
-import { SendIcon } from '../icons/SendIcon'
-
 export const navItems = [
   {
     title: 'Auctions',
     url: '/overview#topAnchor',
-    icon: <SendIcon />,
```
josojo, line 9: hehe

elena-zh:
Hi @gabitoesmiapodo , I have verified the changes.

So, almost all the cases are fixed. However, here are my notes: 
1. sentences in a private auction's text have different alignment (Android): 


2. Still: Iphone 12mini (I do not have other iPhone devices to check): auction details are missing at all now (any auction): 


3. Chart is not fixed and it is too small in iPhone 12 mini, but, as you have explained, it might be a device-specific issue...

4. As for the Back button, I will create a low prio enhancement for this case

PR #445, opened 2021-04-07
(draft) Re enable Wallet Connect.

gabitoesmiapodo:
Regarding the latest discussions about re enabling WalletConnect:

- I created this pull request for everyone to test this.
- I added a white border to the QR code, that makes scanning it easier.
- I opened Gnosis Auction from my desktop browser, and used the WalletConnect compatible Metamask app (but any other compatible app should be the same). I could connect and had no issues.
- Metamask must be disabled on the desktop browser, otherwise it overrides WalletConnect's connection by using Metamask by default. This is pretty annoying, and possibly the main issue to fix. 
- If Metamask is absent / disabled / locked, there are no issues when connecting, or otherwise.
- Placed a few different orders (USDC, DAI, etc.), and had no issues there. All the confirmations were requested on my phone, and when confirmed the computer's browser reflected that.
- All my tests where performed on Rinkeby.
- It obviously makes not sense to have the WalletConnect option on mobile, but it's there. Maybe we can hide it.

So guys please have a look and see if this works for you, and if there's anything else you can do about these issues.

### DIFF
--- a/src/components/modals/common/WalletConnectData/index.tsx
+++ b/src/components/modals/common/WalletConnectData/index.tsx
@@ -11,6 +11,10 @@ const Wrapper = styled.div`
   margin-bottom: 20px;
 `
 
+const QR = styled(QRCode)`
+  border: 10px solid #fff;
+`
+
 interface WalletConnectDataProps {
   size: number
   uri?: string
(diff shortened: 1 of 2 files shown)


jarcodallo:
I couldn't open the QR in brave w/o metamask. I pressed the button multiple times and nothing happened. Only increases the `warn.js:2 [react-ga] ReactGA.initialize must be called first or GoogleAnalytics should be loaded manually` warning. 

image

Same problem in chrome:
image

jarcodallo:
if metamask is installed works fine, but if I reload the page the walletconnect address disappears.

image

after reload the page
image

gabitoesmiapodo:
Fixed on https://github.com/gnosis/ido-ux/pull/447 so I'm closing this.

PR #447, opened 2021-04-07
Bug: Walletconnect disconnects after page reload

mikheevm:
So the problem was that the `useEagerConnect` hook only checks if there's an existing metamask connection, it doesn't check walletconnect. 

This PR:
- Add walletconnect check to useEagerConnect
- Increase supported node versions in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "homepage": ".",
   "private": true,
   "engines": {
-    "node": ">=4.4.7 <=14.15.4"
+    "node": ">=4.4.7 <=14.16.4"
   },
   "scripts": {
     "start": "react-scripts start",
--- a/src/components/common/UserDropdown/index.tsx
+++ b/src/components/common/UserDropdown/index.tsx
@@ -191,6 +191,7 @@ export const UserDropdown: React.FC<Props> = (props) => {
 
   const disconnect = React.useCallback(async () => {
     deactivate()
+    localStorage.removeItem('walletconnect')
   }, [deactivate])
 
   const UserDropdownContent = () => {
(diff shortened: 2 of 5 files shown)


josojo (approved):
I tested it throughly and I think we should ship it asap!

elena-zh:
HI @mikheevm , 
connection works pretty cool with Rinkeby: I was able to connect/disconnect a wallet, place/cancel/claim an order, change accounts and etc.
The issue appeared when I switched to the Mainnet in the connected Mobile Metamask:

And then I'm not able to connect a wallet using walletconnect even if I change network to Rinkeby in my mobile metamask.
Could you please take a look into this issue? 
Thanks

elena-zh:
The same issue appears when  switch to xDAI in the connected wallet

elena-zh:
The issue described [here ](https://github.com/gnosis/ido-ux/pull/445#issuecomment-814729751) is also reproducible in this PR.

mikheevm:
@elena-zh why is this an issue? the deployed version is meant to be used with rinkeby, it shouldn't work for other networks

elena-zh:
@mikheevm , 
Production version should work with the Mainnet and xDAI only. 
Rinkeby should not be supported there. 
That is why it is an issue. IMHO

mikheevm:
@elena-zh it's not an easy fix with the current implementation, the interface depends on REACT_APP_CHAIN_ID environment variable: https://github.com/gnosis/ido-ux/blob/develop/.env.example#L8

for the production it's set to 1 (mainnet), thus it will only work on mainnet

PR #452, opened 2021-04-07
inserting the logo again

josojo:
I think in the other PR is was deleted by accident.

### DIFF
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -295,6 +295,13 @@ const AuctionDetails = (props: AuctionDetailsProps) => {
         itemValue={
           derivedAuctionInfo?.auctioningToken && derivedAuctionInfo?.initialAuctionOrder ? (
             <>
+              <TokenLogo
+                size={'20px'}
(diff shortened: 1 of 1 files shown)


gabitoesmiapodo:
> I think in the other PR is was deleted by accident.

Nope. The logo is not there in the design, actually: https://www.figma.com/file/hapgMbgd2Zo5NFZDNfbogQ/Gnosis-Auction-Feedback?node-id=1%3A3

And it looks like this in the mobile version:

2021-04-07 11_45_10-Gnosis Auction

josojo:
@gabitoesmiapodo 

you are right and sorry that I just merged this sooo quickly.

But it was important for us. In the boson sale guide it was mentioned that people should always look for the logo to identify the correct auction. If they wouldn't see the logo, the guide told them that it would be scam. :) 

Sure, you could not know this, but because of this, I inserted it quickly again.

My fix actually broke the mobile design. :) Sorry for that. If you could look into a fix with the logo, this would be great.
Sorry again

gabitoesmiapodo:
Sure, I'll fix it, no problem.

PR #459, opened 2021-04-07
Selling volume incorrect

### LINKED ISSUE #402: Selling volume incorrect
It should be 0.2*2000 = 400 USDT, and not 0.2 USDT

mariano-aguero:
Closes #402


josojo:
Yes, the result is correct!

Though, it should be easier to get. You could have just used the yValue of the one ask instead of introducing a new variable and maintain it per order

mariano-aguero:
> Yes, the result is correct!
> 
> Though, it should be easier to get. You could have just used the yValue of the one ask instead of introducing a new variable and maintain it per order

Mmmm I will go with another approach, I was concerned about decimals, and because the initial setup of the property askY  were null. I added an adapter to create the tooltip using that property, and checking that it is not null. Also I added a toFixed.

josojo (approved):
Nice Great!

Much better solution than before.

I just changed the precision such that auctions selling against WETH with decimals as 0.001 are also well displayed

mariano-aguero:
@josojo I saw that the branch development is the default, and possibly being used in production, so just in case, I leave the option of merge this against development in your hands

josojo:
@josojo I saw that the branch development is the default, and possibly being used in production, so just in case, I leave the option of merge this against development in your hands

Thanks. but no worries, we don't do auto deployments to production.

PR #460, opened 2021-04-07
(add) "Accept Selected" button

### LINKED ISSUE #448: Cookies banner not correct
Legal complained that we need to offer the users an option to really only accept the necessary cookies. It should look like this:

### LINKED ISSUE #449: Change cookies banner and allow cookies selection
Our legal team informed me we need to improve our cookies handling.
We should allow the user to select

Here's an example of the cookies types

LMK if you need any further clarification from legal

gabitoesmiapodo:
Closes #448 
Closes #449


josojo:
Since this is also a bigger change, I would prefer to merge it after the boson sale.

gabitoesmiapodo:
Sure, I'll add a label so we don't forget.

gabitoesmiapodo:
This feature is dead now, so I'm closing this.

Issue #464, opened 2021-04-07
Chart height on iPhone


gabitoesmiapodo:
See `Item #3` on #463

PR #465, opened 2021-04-07
(Fix) Mobile issues

### LINKED ISSUE #463: Some more mobile issues
Sentences in a private auction's text have different alignment (Android): 

2. Still: Iphone 12mini (I do not have other iPhone devices to check): auction details are missing at all now (any auction): 

3. Chart is not fixed and it is too small in iPhone 12 mini, but, as you have explained, it might be a device-specific issue...

_Originally posted by @elena-zh in https://github.com/gnosis/ido-ux/issues/444#issuecomment-814868605_

gabitoesmiapodo:
Closes #463

(1) Should be fixed.
(2) I suspect it was a temporary issue (can't be sure, sorry).
(3) I asked a friend who owns an iPhone 8 and it has the same issue with the chart's height. I created an issue here https://github.com/gnosis/ido-ux/issues/464 to try to fix that.

### DIFF
--- a/src/components/auction/Claimer/index.tsx
+++ b/src/components/auction/Claimer/index.tsx
@@ -23,6 +23,7 @@ const Wrapper = styled(BaseCard)`
 `
 
 const ActionButton = styled(Button)`
+  flex-shrink: 0;
   height: 52px;
   margin-top: auto;
 `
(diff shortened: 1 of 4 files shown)


jarcodallo (approved):
🚀

elena-zh:
@gabitoesmiapodo , 
case 1 is fixed, 
case 2 is still reproducible. I created a low priority issue for this case #477

PR #470, opened 2021-04-08
fix: Always allows a user to connect with wallet connect

josojo:
Currently, we have an issue that some users that don't have metamask installed, can't connect with wallet connect.

This fix removes the unnecessary check for the networkID - that can't work - if metamask is not connected.

-----
I don't understand it fully, but I think its the right fix. Please review carefully.


elena-zh (approved):
HI @josojo , 

Great job!
I was able to connect a wallet using wallet connect in different browsers (Brave, FF, Chrome) when Metamask was not installed/switched off/switched on/locked.
And then I was able to place/cancel orders using a connected wallet via WalletConnect.

josojo:
For code comparison, see:
https://github.com/Uniswap/uniswap-interface/blob/8fd894f2d1b03921c5519c050aed164343c47fb1/src/components/WalletModal/index.tsx#L188

PR #471, opened 2021-04-08
Does not allow to place orders below the clearing price

josojo:
Users were placing orders bleow the clearing price, although this makes no sense and they are just wasting gas.

Hence, in the no-order-cancelation phase, we don't allow placing orders below the current clearing price, as the clearing price can only increase.

can be easily tested on auction: `auction?auctionId=33&chainId=4#topAnchor`

### DIFF
--- a/src/components/auction/OrderPlacement/index.tsx
+++ b/src/components/auction/OrderPlacement/index.tsx
@@ -148,7 +148,7 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
   const orders: OrderState | undefined = useOrderState()
   const toggleWalletModal = useWalletModalToggle()
   const { chainId: auctionChainId, price, sellAmount } = useSwapState()
-  const { error } = useGetOrderPlacementError(derivedAuctionInfo)
+  const { error } = useGetOrderPlacementError(derivedAuctionInfo, auctionState)
   const { onUserSellAmountInput } = useSwapActionHandlers()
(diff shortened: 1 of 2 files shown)


elena-zh:
Hi @josojo , 
I have verified the fix. 
Works pretty good, however, here are my notes: 
1. It looks a bit strange when I see 1st limitation (min price, I guess), then I see the next one (less than clearing price). It would be nice to show only one limitation amount, I think
https://drive.google.com/file/d/1xyaXJv-kPQAAT6FTok67v_I9Hwz7U5c0/view

2. Then, validation message disappears when I enter an amount < than clearing price, but with a lot of '9' decimals (it might be a rounding issue). And I'm able to run this transaction


3. It would be nice to set clearing price into the Price field instead of showing a value = 'min price+1%'


Let me know if I need to create separate issues for these cases.

josojo:
@elena-zh 

nice catches.

I fixed 1, lets make separate issues for 2. and 3.

josojo:
Please test it, although I am already merging it for my ipfs deployment

elena-zh:
@josojo , the changes LGTM. 
I've taken a look into validation messages for

- auction without cancellation date  --> allow to place orders < clearing price;
- auction before cancellation date --> allow to place orders < clearing price;
- auction after cancellation date --> does not allow to place orders < clearing price.

Separate tasks #474 an #475 for cases 2-3 are created.

Thanks

PR #476, opened 2021-04-08
(fix) cursor flickering

gabitoesmiapodo:


josojo (approved):
Thanks for fixing it soo quickly!

PR #480, opened 2021-04-08
(Fix) Waiting modal appears when clicking on the Claim button again

### LINKED ISSUE #359: Waiting modal appears when press on the claim button once again
1. create an order for an auction
2. wait till the auction is settled
3. claim funds, confirm the transaction
4. while transaction is being executed, press on the Claim button once again

**AR**: waiting pop-up appears with a lot of errors in the console

**ER**: disable 'Claim' button once transaction is confirmed in a connected wallet

gabitoesmiapodo:
Closes #359 

This should work better now.

In the future we should modify the "Error" logic to make it capable of dealing with different status returned by the hooks.

We should be able to discern if we are dealing with an actual error, or if we are dealing with a "Success" status (ie: "You already claimed your funds" is an OK status and it should display a green checkmark, and not the red warning icon).

### DIFF
--- a/src/components/auction/Claimer/index.tsx
+++ b/src/components/auction/Claimer/index.tsx
@@ -60,54 +60,56 @@ const Text = styled.div`
   line-height: 1.2;
   margin-left: 10px;
 `
-interface ClaimerProps {
+
+interface Props {
   auctionIdentifier: AuctionIdentifier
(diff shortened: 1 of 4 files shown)


elena-zh:
Changes LGTM!

josojo (approved):
LGTM

PR #488, opened 2021-04-09
(fix) iOS fixes

### LINKED ISSUE #464: Chart height on iPhone

### LINKED ISSUE #477: No auction details in iPhone browsers
Device: iphone 12 mini, Chrome+Safari
Case 2 in #463 
image

I opened the app in incognito tab, I got this issue reproducible.

gabitoesmiapodo:
Closes #464 
Closes #477

All right, someone lent me an iPhone 8 (the closest I could get to an iPhone 12, sorry) and I fixed several things. I could actually reproduce the same problems that were reported on other iOS devices, so I think the fixes here will work there too.

These screenshots are from the iPhone 8 I used:

WhatsApp Image 2021-04-09 at 16 21 21
WhatsApp Image 2021-04-09 at 16 21 44
WhatsApp Image 2021-04-09 at 16 22 09

Android, desktop, etc. should be the same as it was before.

### DIFF
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -31,18 +31,19 @@ const Wrapper = styled(BaseCard)`
   grid-template-columns: 1fr 3px 1fr;
   grid-template-rows: 1fr;
   padding-bottom: 20px;
+  padding-top: 75px;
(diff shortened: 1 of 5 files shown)


josojo (approved):
LGTTM

PR #491, opened 2021-04-15
frist step to inverted prices

josojo:
Just a draft: 
Tries to improve the situation for yfi sale.

Unfortunately, I could not improve the chart, because then we would get performance problems. Chart is too slow

### DIFF
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -7,6 +7,7 @@ import {
   DerivedAuctionInfo,
   orderToPrice,
   orderToSellOrder,
+  useSwapState,
 } from '../../../state/orderPlacement/hooks'
 import { AuctionIdentifier } from '../../../state/orderPlacement/reducer'
 import { getExplorerLink, getTokenDisplay } from '../../../utils'
@@ -194,6 +195,7 @@ const AuctionDetails = (props: Props) => {
     [chainId, derivedAuctionInfo?.biddingToken],
   )
 
+  const { showPriceInverted } = useSwapState()
   const { clearingPriceInfo } = useClearingPriceInfo(auctionIdentifier)
   const biddingTokenDisplay = useMemo(() => getTokenDisplay(derivedAuctionInfo?.biddingToken), [
     derivedAuctionInfo?.biddingToken,
@@ -210,20 +212,38 @@ const AuctionDetails = (props: Props) => {
         derivedAuctionInfo?.biddingToken,
         derivedAuctionInfo?.auctioningToken,
       )
-    const clearingPriceNumber = orderToPrice(clearingPriceInfoAsSellOrder)?.toSignificant(4)
+    const clearingPriceNumber =
+      showPriceInverted && orderToPrice(clearingPriceInfoAsSellOrder)
+        ? orderToPrice(clearingPriceInfoAsSellOrder).invert().toSignificant(5)
+        : orderToPrice(clearingPriceInfoAsSellOrder)?.toSignificant(5)
 
     return clearingPriceNumber ? (
-      <>
-        <TokenValue>{abbreviation(clearingPriceNumber)}</TokenValue>{' '}
-        <TokenSymbol>
-          {getTokenDisplay(derivedAuctionInfo?.biddingToken)} per{' '}
-          {getTokenDisplay(derivedAuctionInfo?.auctioningToken)}
-        </TokenSymbol>
-      </>
+      showPriceInverted ? (
+        <>
+          <TokenValue>{abbreviation(clearingPriceNumber)}</TokenValue>{' '}
+          <TokenSymbol>
+            {getTokenDisplay(derivedAuctionInfo?.auctioningToken)} per{' '}
+            {getTokenDisplay(derivedAuctionInfo?.biddingToken)}
+          </TokenSymbol>
+        </>
+      ) : (
+        <>
+          <TokenValue>{abbreviation(clearingPriceNumber)}</TokenValue>{' '}
+          <TokenSymbol>
+            {getTokenDisplay(derivedAuctionInfo?.biddingToken)} per{' '}
+            {getTokenDisplay(derivedAuctionInfo?.auctioningToken)}
+          </TokenSymbol>
+        </>
+      )
     ) : (
       '-'
     )
-  }, [derivedAuctionInfo?.auctioningToken, derivedAuctionInfo?.biddingToken, clearingPriceInfo])
+  }, [
+    derivedAuctionInfo?.auctioningToken,
+    showPriceInverted,
+    derivedAuctionInfo?.biddingToken,
+    clearingPriceInfo,
+  ])
 
   const titlePrice = useMemo(
     () =>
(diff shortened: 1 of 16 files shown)


src/constants/config.ts
```diff
@@ -27,3 +27,8 @@ export const PORTIS_ID = process.env.REACT_APP_PORTIS_ID || ''
 // Other stuff
 export const GOOGLE_ANALYTICS_ID = process.env.REACT_APP_GOOGLE_ANALYTICS_ID || ''
 export const GIT_COMMIT_HASH = process.env.REACT_APP_GIT_COMMIT_HASH || ''
+
+export const STABLE_TOKEN_ADDRESSES = [
```
josojo, line 31: This is no longer needed, I will remove it

src/utils/prices.ts
```diff
@@ -2,8 +2,19 @@ import { Token } from 'uniswap-xdai-sdk'
 
 import { BigNumber } from '@ethersproject/bignumber'
 
+import { STABLE_TOKEN_ADDRESSES } from '../constants/config'
 import { tryParseAmount } from '../state/orderPlacement/hooks'
 
+export function shouldPricesBeInverted(auctioningTokenAddress: string): boolean {
```
josojo, line 8: I will also remove this funciton, as it is no longer needed

src/components/auction/AuctionDetails/index.tsx
```diff
@@ -210,20 +212,38 @@ const AuctionDetails = (props: Props) => {
         derivedAuctionInfo?.biddingToken,
         derivedAuctionInfo?.auctioningToken,
       )
-    const clearingPriceNumber = orderToPrice(clearingPriceInfoAsSellOrder)?.toSignificant(4)
+    const clearingPriceNumber =
+      showPriceInverted && orderToPrice(clearingPriceInfoAsSellOrder)
+        ? orderToPrice(clearingPriceInfoAsSellOrder).invert().toSignificant(5)
+        : orderToPrice(clearingPriceInfoAsSellOrder)?.toSignificant(5)
```
hpmaxi, line 218: `orderToPrice` seems like is having two responsibilities, checking if that order has a "calculable" price and calculating it. So I think the first call to it is unnecessary because then is used with `?.` operator. Now is possible to have `showPriceInverted` as false and the first call to `orderToprice` not called. Another option is to extract everything as `priceFromOrder` or something similar.

src/components/auction/AuctionDetails/index.tsx
```diff
@@ -210,20 +212,38 @@ const AuctionDetails = (props: Props) => {
...
+        </>
+      ) : (
+        <>
+          <TokenValue>{abbreviation(clearingPriceNumber)}</TokenValue>{' '}
+          <TokenSymbol>
+            {getTokenDisplay(derivedAuctionInfo?.biddingToken)} per{' '}
+            {getTokenDisplay(derivedAuctionInfo?.auctioningToken)}
+          </TokenSymbol>
+        </>
+      )
```
hpmaxi, line 237: I'm bad for names, but looking at this the only thing that changes is the order of the string inside `<TokenSymbol>` repeated twice. Maybe it could be extracted to a single value that depends of showPriceInverted, `useMemo` for example, or a function that receives `derivedAcutionInfo` and if they should be inverted or not.

src/components/auction/OrderbookWidget/index.tsx
```diff
@@ -140,8 +140,8 @@ const processData = (
         // Data for representation
         priceNumber: price,
         totalVolumeNumber: totalVolume,
-        priceFormatted: price.toFixed(2),
-        totalVolumeFormatted: totalVolume.toFixed(2),
+        priceFormatted: price.toFixed(5),
+        totalVolumeFormatted: totalVolume.toFixed(5),
```
hpmaxi, line 144: Not sure about this, but it possible that a constant like `MAX_DECIMALS_PRICE_FORMAT` or similar could clear the intent and make any refactor/change easier to do.

src/components/form/PriceInputPanel/index.tsx
```diff
@@ -26,40 +26,58 @@ interface CurrencyInputPanelProps {
...
+              <DoubleLogo
+                auctioningToken={{
+                  address: biddingToken.address,
+                  symbol: biddingToken.symbol,
+                }}
+                biddingToken={{
+                  address: auctioningToken.address,
+                  symbol: auctioningToken.symbol,
+                }}
+                size="24px"
```
hpmaxi, line 66: `DoubleLogo` could be responsible for ordering them, maybe for another developer in the future, a prop called `biddingToken` receiving an `auctioningToken` can be read as an error.

src/state/orderPlacement/reducer.ts
```diff
@@ -44,6 +48,13 @@ export default createReducer<SwapState>(initialState, (builder) =>
         sellAmount,
       }
     })
+    .addCase(invertPrice, (state) => {
+      return {
+        ...state,
+        price: getInverse(Number(state.price), 16).toString(),
+        showPriceInverted: !state.showPriceInverted,
+      }
```
hpmaxi, line 56: This is the correct place to have this kind of logic but seems repeated a few times across this pull request. `useSwapPrice` can be modified, or a selector for price can be used to avoid that.  Something to take into account at the moment of a refactor.

src/utils/prices.ts
```diff
@@ -2,8 +2,19 @@ import { Token } from 'uniswap-xdai-sdk'
...
+  return STABLE_TOKEN_ADDRESSES.includes(auctioningTokenAddress)
+}
+
+export function getInverse(price: number, digits: number): number {
+  const re = new RegExp('(\\d+\\.\\d{' + digits + '})(\\d)'),
+    m = (1 / price).toString().match(re)
+  return m ? parseFloat(m[1]) : (1 / price).valueOf()
+}
+
```
hpmaxi, line 17: This takes the first matching group of a decimal number with a certain amount of digits after a period and based on that does (1/price) two times (giving back price again) or not?.  Maybe by looking at it more time I can have a better understanding of what it is doing.

Some ideas
- Leave a comment about the two capture groups of the regexp or give them a name, `const [foo, bar] = m` or ?<name> but the latter make it harder to read IMHO.
- Leave a comment about what it means to not have a match, maybe a numeric example (or more than one).
- Some unit testing, will this always work for javascript float and number precision?

hpmaxi (approved):
In general looks good to me, and it also seems to work well. Some comments to improve readability are left.

src/state/orderPlacement/reducer.ts
josojo: It's tricky. Probably, the most tricky part of this PR.

I am actually not sure what the best way is to store the price. Maybe it would have been better to store it always in the canonical "form", - as we will submit it to the contract. 
Here I am storing it, as the user inserts the value, but this forces me to later change the values again for contract submission.

Let me think about it a little bit more and then probably, I will refactor it

src/utils/prices.ts
josojo: Yes, unit tests are the best. I will do it later, as I still want to do the release today

josojo:
Thank you for your good review! @hpmaxi

josojo:
While the PR is not yet great, I am merging it to quickly present something to the yfi community. :)

PR #495, opened 2021-04-16
Native orders

josojo:
This PR enables to place orders with xdai instead of wxdai.

I did not enable this for all networks, as I am not sure how good it would be for the UX. One bad point is that users deposit xdai, but only get wxdai, later. Hence, I am not sure whether we really want to enable that for all networks.


src/utils/index.ts
```diff
@@ -174,6 +174,7 @@ export function escapeRegExp(string: string): string {
 }
 
 // Always return a non-undefined token display
-export function getTokenDisplay(token: Token): string {
+export function getTokenDisplay(token: Token, chainId: ChainId): string {
+  if (token.address == WETH[chainId].address && chainId == 100) return `XDAI`
```
hpmaxi, line 178: Maybe this could be extracted, the condition as a function predicate.

hpmaxi (approved):
Only left one comment about the code. Besides that, everything looks good.

josojo:
closes #492

PR #496, opened 2021-04-17
fix: non-participants don't get a spinner for claiming dialog

josojo:
non-participants of an auction would always get a spinner instead of the claiming dialogue. This PR fixes it.


josojo:
This was just an issue I had with my brave... really strange, it connected to rinkeby all the time. Need more investigation.

PR #514, opened 2021-04-26
(fix) 'Install Metamask' button

### LINKED ISSUE #478: 'Install Metamask' button is enabled when 'Terms' box is unchecked
1. Disable metamask in browser
2. call the connection modal 

**AR**: 'Install Metamask' button is enabled when 'Terms' box is unchecked

**ER**: 'Install Metamask' button should be disabled when 'Terms' box is unchecked

gabitoesmiapodo:
Closes #478

Link removed.

### DIFF
--- a/src/components/modals/WalletModal/index.tsx
+++ b/src/components/modals/WalletModal/index.tsx
@@ -6,7 +6,6 @@ import { URI_AVAILABLE } from '@anxolin/walletconnect-connector'
 import { UnsupportedChainIdError, useWeb3React } from '@web3-react/core'
 import ReactGA from 'react-ga'
 
-import MetamaskIcon from '../../../assets/images/metamask.png'
 import { fortmatic, injected, walletconnect } from '../../../connectors'
 import { OVERLAY_READY } from '../../../connectors/Fortmatic'
 import { SUPPORTED_WALLETS } from '../../../constants'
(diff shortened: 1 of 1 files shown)


src/components/modals/WalletModal/index.tsx
```diff
@@ -195,18 +194,7 @@ const WalletModal: React.FC = () => {
 
       if (option.connector === injected) {
         if (!(window.web3 || window.ethereum)) {
-          if (option.name === 'MetaMask') {
-            return (
-              <Option
-                icon={MetamaskIcon}
-                key={key}
-                onClick={() => window.open('https://metamask.io/')}
```
josojo, line 203: I think providing the link is to metamask is quite useful. I would not want to remove it.

I would prefer to disable the button, until the user agrees to terms and then still continue to show this link.

gabitoesmiapodo: That could end up being kind of confusing: The user has to accept the T&C to enable a button which is a link to install Metamask (accepting the T&C to install an external plugin is weird IMO), which - after installing it- will be the button to connect to Metamask... that's convoluted.

A better (or just simpler) option could be a line of text at the bottom, something like "Don't have a wallet? Try Metamask!" could do it.

josojo: >A better (or just simpler) option could be a line of text at the bottom, something like "Don't have a wallet? Try Metamask!" could do it.

you are right. Probably that is simpler.

hpmaxi (approved):

src/components/modals/WalletModal/index.tsx
gabitoesmiapodo: Should look like this now.

2021-04-27 15_15_14-Gnosis Auction

josojo (approved):
Perfect yes. This is a good solution

Issue #515, opened 2021-04-26
processOrderBookData tests & performance


hpmaxi:
`processOrderbookData` is taking ~20ms for `auctionId=21` in rinkeby, this could be taking much more with a lot of bids. And runs every time the user change the price amount.

To improve this is necessary to make some unit testing for the function and his auxiliary functions too. And then find the bottleneck to try to find a faster way to do the same calculations. 

What do you think @josojo ?

josojo:
>processOrderbookData is taking ~20ms for auctionId=21 in rinkeby, this could be taking much more with a lot of bids. And runs every time the user change the price amount.

This auction has already 5000 orders. Not sure how bad 20ms are actually in this case.

>To improve this is necessary to make some unit testing for the function and his auxiliary functions too. And then find the bottleneck to try to find a faster way to do the same calculations.

Yeah, most likely we are losing time with all the sorting. This seems to be the only expensive function call.


Overall I am not sure whether 20ms are worth the effort. I don't have sufficient experience to judge

PR #516, opened 2021-04-26
Memoize function prop and order book data

hpmaxi:
`CurrencyInputPanel` was rendering multiple times because onMax was changing on each render. Also processData for orderBook is memorized, but it would be better to or denounce it, or make it faster (see #515).

### DIFF
--- a/src/components/auction/OrderPlacement/index.tsx
+++ b/src/components/auction/OrderPlacement/index.tsx
@@ -168,7 +168,16 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
   const [attemptingTxn, setAttemptingTxn] = useState<boolean>(false) // clicked confirmed
   const [pendingConfirmation, setPendingConfirmation] = useState<boolean>(true) // waiting for user confirmation
   const [txHash, setTxHash] = useState<string>('')
-  const parsedBiddingAmount = tryParseAmount(sellAmount, derivedAuctionInfo?.biddingToken)
+
+  const auctioningToken = React.useMemo(() => derivedAuctionInfo.auctioningToken, [
+    derivedAuctionInfo.auctioningToken,
+  ])
+
+  const biddingToken = React.useMemo(() => derivedAuctionInfo.biddingToken, [
+    derivedAuctionInfo.biddingToken,
+  ])
+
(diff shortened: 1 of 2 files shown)


src/components/auction/OrderPlacement/index.tsx
```diff
@@ -380,8 +386,8 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
             {notApproved && (
               <ApprovalWrapper>
                 <ApprovalText>
-                  You need to unlock {derivedAuctionInfo?.biddingToken.symbol} to allow the smart
-                  contract to interact with it. This has to be done for each new token.
+                  You need to unlock {biddingToken.symbol} to allow the smart contract to interact
```
josojo, line 389: SHould it be biddingTokenDisplay instead of biddingToken.symbol?

src/components/auction/OrderPlacement/index.tsx
```diff
@@ -219,12 +219,18 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
     setAttemptingTxn(false)
   }
 
+  const auctioningToken = React.useMemo(() => derivedAuctionInfo?.auctioningToken, [
+    derivedAuctionInfo?.auctioningToken,
+  ])
+
+  const biddingToken = React.useMemo(() => auctioningToken, [auctioningToken])
```
josojo, line 226: Probably, this line is not intended. I don't understand the switch. of bidding and auctioningToken

src/components/auction/OrderPlacement/index.tsx
```diff
@@ -219,12 +219,18 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
     setAttemptingTxn(false)
   }
 
+  const auctioningToken = React.useMemo(() => derivedAuctionInfo?.auctioningToken, [
```
josojo, line 222: If we move these defintions up several lines, we can reuse this definition even more times.

src/components/auction/OrderPlacement/index.tsx
hpmaxi: Yes, good catch!.

josojo (approved):

PR #518, opened 2021-04-27
Update devguide08.md

avsavsavs:


src/docs/files/devguide08.md
```diff
@@ -0,0 +1,62 @@
+### How to Participate as an Auctioneer using the Gnosis Safe App
+
+#### Connect Wallet & App
+
+1. First, you will need to have a [Gnosis Safe](https://gnosis-safe.io/) that you control deployed in Mainnet.
```
josojo, line 5: I think we should mention that the app is also available on rinkeby. But this guide focuses on mainnet

src/pages/Documentation/index.tsx
```diff
@@ -198,10 +201,13 @@ export const Documentation: React.FC = (props) => {
             Participate as a bidder
           </IndexLink>
           <IndexLink activeClassName="isActive" to="/docs/participate-as-auctioneer#topAnchor">
-            Participate as auctioneer
+            Start an auction via cmd
```
josojo, line 204: I renamed it for a shorter naming and more similar to the       
      Start an auction via UI

Participating wasn't the right word in my opinion

src/docs/files/devguide08.md
```diff
@@ -0,0 +1,62 @@
+### How to Participate as an Auctioneer using the Gnosis Safe App
```
josojo, line 1: This is soo long and pariticpate is also not the right word in my opinion

PR #519, opened 2021-04-27
Docs improvements

josojo:
Puh the quality of our docs is not optimal. This is just a quick prove reading from my side and I found many small issues.

We need to improve it quite a bit!
@Rafanator @avsavsavs Could you also go through it one more time ?

### DIFF
--- a/src/docs/files/devguide06.md
+++ b/src/docs/files/devguide06.md
@@ -30,7 +30,7 @@ Additionally, Auctioneers can further customize the auction by using the followi
 
 **&quot;--is-atomic-closure-allowed&quot;** This parameter enables users to close the auction atomically and submit a final bid by calling the `settleAuctionAtomically` function in the smart contract once the `auction-end-date` has been reached. The auctioneer determines whether this parameter is on or off by specifying True or False.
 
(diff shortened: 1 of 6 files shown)


src/docs/files/devguide06.md
```diff
@@ -28,9 +28,9 @@ Additionally, Auctioneers can further customize the auction by using the followi
...
+**&quot;--is-atomic-closure-allowed&quot;** This parameter enables users to close the auction atomically and submit a final bid by calling the `settleAuctionAtomically` function in the smart contract once the `auction-end-date` has been reached. The auctioneer determines whether this parameter is on or off by specifying True or False.
 
-**&quot;--allow-list-manager&quot;** If whitelisting of addresses is required, the auctioneer can provide his/her own allowlistManager contract, or use the AllowListOffChainManaged.json file provided in [this repo](https://github.com/gnosis/ido-contracts/tree/main/deployments). The command needs to be followed by the contract address of the allowlistManager being used. 
+**&quot;--allow-list-manager&quot;** If whitelisting of addresses is required, the auctioneer can provide his/her own allowlistManager contract, or use the AllowListOffChainManaged.json file provided in [this repo](https://github.com/gnosis/ido-contracts/tree/main/deployments). The command needs to be followed by the contract address of the allowlistManager being used.
```
avsavsavs, line 33: instead of using "If whitelisting of address is required" use "If an Allow List of addresses is required, where only approved users will be able to participate, the auctioneer..."

PR #522, opened 2021-04-29
Update devguide06.md

avsavsavs:


src/docs/files/devguide06.md
```diff
@@ -82,3 +82,16 @@ An example of a more complex auction:
...
+
+### Allow-Listing: Generating signatures
+Signatures for an auction with participation restriction can be created like that:
+1. Create a file: `your_address_inputs.txt` with comma separated addresses that should be allow-listed for the auction
+2. Initiate the auction and remember your auctionId
+3. Run the following script:
+```
+export NETWORK=<Your Network>
+export INFURA_KEY=<Your infura key>
+export PK=<Your private key _for the signing address_. The address for this key should not hold any ETH>
```
josojo, line 94: I think we should write:
```
the address is not required to hold any ETH
```
but it can.

src/docs/files/devguide06.md
```diff
@@ -82,3 +82,16 @@ An example of a more complex auction:
...
+yarn hardhat generateSignatures --auction-id "Your auctionId" --file-with-address "./your_address_inputs.txt" --network $NETWORK
```
josojo, line 95: to be uniform, we should write: <Your auctitonId>

PR #529, opened 2021-04-29
Remove chainId env var

hpmaxi:
Shows message to switch networks instead of balance.


src/connectors/index.ts
```diff
@@ -18,10 +17,11 @@ interface NetworkConnectorArguments {
 }
 
 const POLLING_INTERVAL = 10000
+const DEFAULT_CHAIN_ID = 1
 
 const networkConnectorArguments: NetworkConnectorArguments = {
   urls: [],
-  defaultChainId: CHAIN_ID,
+  defaultChainId: DEFAULT_CHAIN_ID,
```
josojo, line 24: Why is the default needed?

If it is really needed, I would prefer to check which one of the following envs is set:
```
if (NETWORK_URL_MAINNET)
if (NETWORK_URL_RINKEBY) 
if (NETWORK_URL_XDAI) 
```
and then use the corresponding network Id, instead of hardcoding a solution

src/components/auction/OrderPlacement/index.tsx
```diff
@@ -295,6 +295,17 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
     maxAmountInput && onUserSellAmountInput(maxAmountInput.toExact())
   }, [maxAmountInput, onUserSellAmountInput])
 
+  const balanceString = React.useMemo(() => {
+    return account
+      ? chainId !== chainIdFromWeb3
+        ? 'Switch network'
+        : `${biddingTokenBalance?.toSignificant(6) || '0'} ${getTokenDisplay(
```
josojo, line 302: If chainId !== chainIdFromWeb3 holds, we should not even try to get the balance in order to avoid the web3 error messages

josojo: But if you wanna do it in a separate PR later, that is fine as well

src/connectors/index.ts
hpmaxi: This should be fixed now. Maybe show some kind of error/message when any URL is defined.

src/components/auction/OrderPlacement/index.tsx
hpmaxi: This shouldn't fetch the balance using `chainId`, seems like every call composes with `useActiveWeb3React` so the single source of truth of which chainId continues to be the provider. 

Anyways if something extra must be changed when `chainId !== chainIdFromWeb3` I will do it in another PR.

josojo (approved):
Thanks

PR #533, opened 2021-05-03
Add PENDING_CANCELLATION state #260

### LINKED ISSUE #260: Add "pending cancellation" state
When a new order is submitted, we have a "pending" status until the tx is mined
When an order is cancelled, the order disappears immediately without getting tx confirmation. 
We should add a "pending cancellation" status for orders that are pending cancellation but tx wasn't mined yet.

henrypalacios:
Closes #260 
Added a status of `PENDING_CANCELLATION`

When an order is cancelled, disappeared immediately using the action [`RemoveOrders`](https://github.com/gnosis/ido-ux/blob/fec72ef5156b4dafb6ecfc5c70945ab4a6a19629/src/state/orders/actions.ts#L13) .

Adding a new status [can be set individually](https://github.com/henrypalacios/ido-ux/blob/ad51de32c7cf51c11b041425360909cd9528200e/src/state/orders/reducer.ts#L51), and after confirmation [it will be removed](https://github.com/henrypalacios/ido-ux/blob/ad51de32c7cf51c11b041425360909cd9528200e/src/state/transactions/updater.tsx#L54).

I have also disabled the cancel button once it has been pressed.
Screenshot from 2021-05-03 15-06-32

Issue: [260](https://github.com/gnosis/ido-ux/issues/260)
_auctionId_: 17
_chainId_: 4

### DIFF
--- a/src/components/auction/OrdersTable/index.tsx
+++ b/src/components/auction/OrdersTable/index.tsx
@@ -147,7 +147,7 @@ const OrderTable: React.FC<OrderTableProps> = (props) => {
   const orderStatusText = {
     [OrderStatus.PLACED]: 'Placed',
     [OrderStatus.PENDING]: 'Pending',
-    [OrderStatus.PENDING_CANCELLATION]: 'Pending-Cancellation',
+    [OrderStatus.PENDING_CANCELLATION]: 'Cancelling',
   }
   const now = Math.trunc(Date.now())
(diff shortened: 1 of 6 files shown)


hpmaxi (approved):

josojo (approved):
Thanks a lot for the valuable contribution

PR #536, opened 2021-05-04
Check account for mismatch of chainIds

### LINKED ISSUE #511: Display of wrong network
No matter on which network I am, I always get the following display while connecting:

I tested it with the auction 
auction?auctionId=18&chainId=1#topAnchor

hpmaxi:
Closes #511 

Add an extra condition, when all accounts are deactivated in metamask, `useActiveWeb3React` returns `chainId = 1`


josojo (approved):
Nice!

PR #538, opened 2021-05-05
rename swap -> orderplacement

josojo:
Simple renaming for better readabilty


hpmaxi (approved):
Looks good. The old name was a little misleading.

PR #539, opened 2021-05-05
auction state handover

josojo:
removing unnecessary auctionState from interfaces of components


hpmaxi (approved):
Looks simpler now!
A minor improvement could be using `const { auctionState } = derivedAuctionInfo` to maintain uniformity. Another option is something like `const { auctionIdentifier, derivedAuctionInfo: { auctionState }, derivedAuctionInfo } = props`

PR #540, opened 2021-05-05
Always claim wxdai, instead of xdai

josojo:
Since the contracts can not pay out xdai, we should always show to claim wxdai

Eg. test on auction auction?auctionId=21&chainId=100#topAnchor


josojo:
Needed for the release, hence, I will merge it now. Please review anyways

PR #546, opened 2021-05-06
(add) wrap / unwrap buttons

gabitoesmiapodo:
I'm not completely sure about what's needed here, but hopefully this is OK. It's obviously still missing some logic, which I think @josojo will implement.

It should look more or less like these:

Screen Shot 2021-05-06 at 14 28 43
Screen Shot 2021-05-06 at 14 28 57
Screen Shot 2021-05-06 at 14 29 10
Screen Shot 2021-05-06 at 14 29 24

We could create a better workflow for this later, maybe similar to what we implemented on https://github.com/gnosis/conditional-tokens-explorer for Wrapping / Unwrapping (which looks or looked like these other images)

09- CTEAF Positions
16- CTEAF Wrap
17- CTEAF Unwrap


josojo (approved):
IMO ready!

PR #547, opened 2021-05-06
Use close if is available

### LINKED ISSUE #535: Wallet connect - disconnect not working
The session end signal is currently not sent properly to the wallets.

Current behaviour:
connect with wallet connect to gnosis-auction.eth
click in the interface "disconnect"
and observe that the session is still "live" in the wallet that was used for wallet connect

Expected behaviour:
connect with wallet connect to gnosis-auction.eth
click in the interface "disconnect"
wallet no longer has an open connection to gnosis-auction.eth.

Can be for example observed with Minerowallet

hpmaxi:
Closes #535

### DIFF
--- a/src/components/common/UserDropdown/index.tsx
+++ b/src/components/common/UserDropdown/index.tsx
@@ -174,7 +174,7 @@ export const UserDropdown: React.FC<Props> = (props) => {
   const [transactionsModalVisible, setTransactionsModalVisible] = React.useState(false)
   const [darkMode, toggleDarkMode] = useDarkModeManager()
 
-  const { deactivate, library } = useActiveWeb3React()
+  const { connector, deactivate, library } = useActiveWeb3React()
 
   const getWalletName = React.useCallback((): string => {
     const provider = library?.provider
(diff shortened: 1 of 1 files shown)


josojo:
Can you please give us reviewers a link to the documentation from which you got that? Or maybe the link to uniswap, if you copied it from there?
I would love to understand thee closing as well

src/components/common/UserDropdown/index.tsx
```diff
@@ -236,7 +236,9 @@ export const UserDropdown: React.FC<Props> = (props) => {
           <DisconnectButton
             buttonType={ButtonType.danger}
             onClick={() => {
-              disconnect()
+              const c: any = connector
+              if (typeof c.close === 'function') c.close()
```
josojo, line 240: Do you understand in which cases the connector does not have this function close?
It seems uniswap always relies on it:
https://github.com/Uniswap/uniswap-interface/blob/09f30ce0f7043399469644ba80763d636ae5a503/src/components/AccountDetails/index.tsx#L309

josojo (approved):
Works smoothly!

src/components/common/UserDropdown/index.tsx
hpmaxi: @josojo 
It depends on the provider, injected provided doesn't have that method. The type is `AbstractConnector`, this is extended and for WalletConnectProvider is available. `AbstractConnector` always define a `deactivate` method. 

In uniswap, for injected/metamask I don't get the disconnect button.

josojo: I like it better to have also a disconnect button for injected providers... 
I don't understand why one would not want to have it. Do you?

PR #552, opened 2021-05-07
(fix) Chart height changing

### LINKED ISSUE #530: Graph height moves with new input
Imagine the following situation in the auction:
https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=69&chainId=4#topAnchor
Amount input: 300
Price input: 278

If we know change the price to a reason number:
Price input: 2780

the whole picture is changing. Especially disturbing is that the height of the graph or/and input box is changed.
Please fix that the height is constant and does not change on new inputs.

gabitoesmiapodo:
Closes #530


josojo (approved):
Thanks

it still changes height, if a new price-input introduces a change of number of "error messages", but this will be fixed with the new approval workflow laters... so all good!

PR #558, opened 2021-05-11
docs: rewriting ### Gnosis Auction User flow

josojo:
I like it much better if it gives a very high-level overview. Personally, in the old version, I would have stopped reading already once the auctioneer parameters are mentioned. They simply do not fit into the user flow for most users

### DIFF
--- a/src/docs/files/devguide04.md
+++ b/src/docs/files/devguide04.md
@@ -1,4 +1,4 @@
-### Gnosis Auction User flow
+### Gnosis Auction user flow
 
 In Gnosis Auction, there are two types of participants: one seller and multiple bidders.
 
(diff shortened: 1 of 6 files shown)


avsavsavs:
Everything looking great!!

src/docs/files/devguide04.md
```diff
@@ -1,35 +1,34 @@
...
+#### 3. Auctioneer: On-chain price calculation
 
-_Important: as a bidder, it is important to note that many participants might submit new bids right before the auction closes, making it difficult to predict the closing price of the auction ahead of time. This, it is important for bidders to submit their bids with the highest price they are comfortable to pay for the asset_
+It's the auctioneers duty to finish the auction by sending the transaction that calculates the price of the auction on-chain. Though it's also a permissionless process and everyone can do it.
+
+Check out the [guide](https://github.com/gnosis/ido-contracts#settle-auctionss) for doing the price calculation.
+
+#### 4. Bidders: Claiming
+
+After the price has settled, the auctioneer will receive the proceeds of the auction. In contracts the bidder will need to submit an additional transaction to claim their auction proceeds.
```
josojo, line 34: itt should be
``` 
In contrast, ..

```

cmagan: Is there a way to make the images clickable so people can enlarge them? For the images are unusable, they are way too small to read

cmagan: Aside from that everything looks good. 
@josojo I liked the whitelist section

cmagan (commented):
Changed a bit the text in the Overview section
Everything looks good

src/docs/files/devguide04.md
josojo: Yes, I agree that htis is a necessary feature. I will create an issue

Issue #507, opened 2021-04-21
JSON RPC error when trying to cancel


cmagan:
The issue is actually that once the cancellation period is over, users will still have the "cancel" button enabled.
Only a refresh will make it disabled.
So if the user tries to cancel after cancellations are not allowed anymore they will get
![image.png](https://images.zenhubusercontent.com/60227f03bd43d8095c1185d0/95e36b10-d021-4c15-90b0-5c9ea3117c95)

PR #563, opened 2021-05-12
Add update AuctionState after order cancellation time expires

henrypalacios:
closes: #507 

When auction `orderCancellationtime` expires, The `auctionState` is not updated so the "cancel" **button remains enable**.

Update a `useEffect` that updates the `auctionState` when the `endTime` expires

⚠️ it seems questionable to me whether to use an **action** through redux, if this event needs to be heard in another component.
- Enviroment
auctionId: 20
chainId: 4

### DIFF
--- a/src/state/orderPlacement/hooks.ts
+++ b/src/state/orderPlacement/hooks.ts
@@ -367,17 +367,28 @@ export function useDerivedAuctionInfo(
 
     const getCurrentState = () => deriveAuctionState(auctionDetails, clearingPriceInfo).auctionState
     setAuctionState(getCurrentState())
-    const timeLeft = calculateTimeLeft(auctionDetails.endTimeTimestamp)
+    const timeLeftEndAuction = calculateTimeLeft(auctionDetails.endTimeTimestamp)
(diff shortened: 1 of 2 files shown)


src/state/orderPlacement/hooks.ts
```diff
@@ -367,17 +367,28 @@ export function useDerivedAuctionInfo(
...
+      const timersId = remainingAuctionTimes
+        .map((timeLeft) => {
+          if (timeLeft < 0) {
+            return
+          }
+          return setTimeout(() => {
+            setAuctionState(getCurrentState())
+          }, timeLeft * 1000)
+        })
+        .filter(isTimeout)
```
josojo, line 385: Like that filtering. Cool.

josojo (approved):
Nice, very clean solution

src/state/orderPlacement/hooks.ts
henrypalacios: Credits to @hpmaxi, who show me the TS "[type-predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)"

PR #564, opened 2021-05-12
Polygon support

### LINKED ISSUE #557: Integrate Polygon network
I will provide the env variables from the backend. 

Contracts can be found here:
https://forum.gnosis.io/t/gip-6-deploy-gnosis-auction/1078/10

hpmaxi:
Closes #557 

This is still WIP, I just want to make it very straightforward to add another network just changing `NETWORK_CONFIG` object.


src/utils/index.ts
```diff
@@ -191,5 +233,5 @@ export function getTokenDisplay(token: Token, chainId: ChainId): string {
 }
 
 export function isTokenXDAI(tokenAddress?: string, chainId?: ChainId): boolean {
-  return !!tokenAddress && !!chainId && tokenAddress == WETH[chainId].address && chainId == 100
+  return !!tokenAddress && !!chainId && chainId == 100 && tokenAddress == WETH[100].address
```
hpmaxi, line 236: This was giving an error with WETH coming from a  library with no support for polygon, because before the number of the chain is also checked there was no need to pass `WETH[chainId]`

src/utils/setupNetwork.ts
```diff
@@ -20,6 +22,27 @@ export const setupNetwork = async (chainId: number) => {
...
+          method: 'wallet_addEthereumChain',
+          params: [
+            {
+              chainId: `0x${chainId.toString(16)}`,
+              chainName: config.name,
+              nativeCurrency: {
+                name: config.name,
+                symbol: config.symbol,
+                decimals: 18,
+              },
```
hpmaxi, line 38: Maybe this could be added to `NETWORK_CONFIG` and just take all the data from there.

josojo (commented):
Really like the spirit of the PR! right approach to try to do it more generic.

I think you missed the wallet connect config for polygon.

Backend should be ready by today and then we can test it

src/utils/index.ts
josojo: For this, we need to clone the library and publish our own one. Could you take care of it?

josojo: Or maybe there is already a clone that works well with xdai polygon and bsc

hpmaxi: @josojo Tried with this fork https://github.com/protofire/uniswap-sdk/tree/support-bsc-polygon but for some reason I can't make it work with the time left.

josojo: Hm... why didn't you fork from the working lib here:
https://github.com/gnosis/ido-ux/blob/develop/package.json#L143
?

.github/workflows/deploy.yml
```diff
@@ -52,8 +52,8 @@ jobs:
           REACT_APP_ADDITIONAL_SERVICES_API_URL_XDAI: https://ido-v1-api-xdai.dev.gnosisdev.com/
           REACT_APP_NETWORK_URL_MAINNET: https://mainnet.infura.io/v3/${{ secrets.INFURA_PROJECT_KEY_DEV }}
           REACT_APP_ADDITIONAL_SERVICES_API_URL_MAINNET: https://ido-v1-api-mainnet.dev.gnosisdev.com/
-          REACT_APP_NETWORK_URL_POLYGON: https://rpc-mainnet.matic.network
-          REACT_APP_ADDITIONAL_SERVICES_API_URL_POLYGON: https://ido-v1-api-polygon.dev.gnosisdev.com/
+          REACT_APP_NETWORK_URL_MATIC: https://rpc-mainnet.matic.network
```
henrypalacios, line 55: @josojo Out of curiosity, why MATIC? if MATIC is just the native token.

PR #567, opened 2021-05-17
[fix]remove function declaration from types.d.ts file

josojo:
Currently develop is no longer working, because in develop `isTimeout` was not defined


josojo:
Merging quickly as an customer is also about to test it...

PR #568, opened 2021-05-17
(fix) isTimeout not found

gabitoesmiapodo:
App was crashing, I think this should fix it.

No related issue.


gabitoesmiapodo:
@josojo fixed this in https://github.com/gnosis/ido-ux/pull/567 so I'm closing this.

Issue #560, opened 2021-05-11
Has pending claim


josojo:
Currently, we are only looking whether a user has a pending approval, but we should do the same thing for claim txs. 
Some users don't understand ethereum logic too well and hence started the claiming very often. E.g this user.

https://etherscan.io/tx/0xafd215353d3f4d1111e653a7ee99398d261dc90c98a7eb5fc7d86d339d30c95c

Expected behavior:

- user start claim processes
- show that the tx in progress, if they started the tx ( similar as approval): (Button should be disabled and should show "claiming", exactly as as "approving- button"

henrypalacios:
Hi @josojo, I have not been to reproduce this.

Let me tell you that I tried.

1. I have created a new Auction and after it is finished, I press the "claim" button then  I could not make a new claim.
![Screenshot from 2021-05-13 15-20-14](https://user-images.githubusercontent.com/4270166/118188791-51874880-b417-11eb-9d7c-573fbd06cefb.png)


2. I though it might only happen at a private auction so I tried to create one but I couldn't get past "**Private Auction**" interface, even though executing this **command** tells me its is done.
```bash 
yarn hardhat generateSignatures --auction-id "87" --file-with-address "./Address.no-commit.txt" --network rinkeby --post-to-api
```

📝 `Address.no-commit.txt` had this:
`0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9,0x1F8D2668BAaB2B324f61fCf34730BE8e79e8eCCF`

josojo:
It's about the situation, where the pending claim tx is in flight. That means the tx was already approved in the wallet, but not yet mined. It's hard to reproduce on test networks, as here, tx are nearly always mined instandely. But on mainnet, mining takes a long time, especially, if a user chooses low gas prices.

PR #571, opened 2021-05-17
Add PENDING ClaimState after claiming

henrypalacios:
closes: #560 

Added `ClaimState`, specifically PENDING status after claim txs.
Selection_283

**Problem:** 
after claiming an auction the user can press the **claim button again**, without waiting for the transaction to be mined.

**Proposal:**
Add transaction detail [in the summary](https://github.com/gnosis/ido-ux/blob/8d0f81033341b12c7d6d7603889033590288a2b1/src/state/transactions/hooks.tsx#L147) to identify **recent unconfirmed transactions** 

⚠️ this solution will not work if i open the auction from another client

**Enviroment:**
ChainId: 4
AuctionId: 89, 90

### DIFF
--- a/src/hooks/useClaimOrderCallback.ts
+++ b/src/hooks/useClaimOrderCallback.ts
@@ -202,17 +202,17 @@ export const useClaimOrderCallback = (
 
   const claimableOrders = claimInfo?.sellOrdersFormUser
   const userHasAvailableClaim = useUserHasAvailableClaim(auctionIdentifier, claimableOrders)
-  const pendingApproval = useHasPendingClaim(auctionIdentifier.auctionId, account)
+  const pendingClaim = useHasPendingClaim(auctionIdentifier.auctionId, account)
 
   const claimStatus = useMemo(() => {
     if (!claimableOrders) return ClaimState.UNKNOWN
 
     return userHasAvailableClaim
-      ? pendingApproval
+      ? pendingClaim
         ? ClaimState.PENDING
         : ClaimState.NOT_CLAIMED
       : ClaimState.CLAIMED
-  }, [claimableOrders, pendingApproval, userHasAvailableClaim])
+  }, [claimableOrders, pendingClaim, userHasAvailableClaim])
 
   return [claimStatus, claimCallback]
 }
--- a/src/hooks/useClaimOrderCallback.ts
+++ b/src/hooks/useClaimOrderCallback.ts
@@ -1,12 +1,13 @@
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import { TokenAmount } from 'uniswap-xdai-sdk'
 
 import { BigNumber } from '@ethersproject/bignumber'
 import { Contract } from '@ethersproject/contracts'
 
+import { useSingleCallResult } from '../state/multicall/hooks'
 import { DerivedAuctionInfo } from '../state/orderPlacement/hooks'
 import { AuctionIdentifier } from '../state/orderPlacement/reducer'
-import { useTransactionAdder } from '../state/transactions/hooks'
+import { useHasPendingClaim, useTransactionAdder } from '../state/transactions/hooks'
 import { ChainId, calculateGasMargin, getEasyAuctionContract } from '../utils'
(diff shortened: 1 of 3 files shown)


henrypalacios:
In Addition, I believe that the responsibility for consulting `claimableOrders` could be removed from[ `useDerivedClaimInfo`](https://github.com/gnosis/ido-ux/blob/7ef6b01b07d79435422e85b872fe3c3a1c957684/src/state/orderPlacement/hooks.ts#L557-L572).

This hook can receive the `ClaimStatus` and display the corresponding error.

src/state/transactions/hooks.tsx
```diff
@@ -129,3 +129,24 @@ export function useHasPendingApproval(tokenAddress?: string, spender?: string):
...
 }
+
+// returns whether a account has a pending claim transaction
+export function useHasPendingClaim(auctionId?: number, from?: string | null): boolean {
+  const allTransactions = useAllTransactions()
+
+  return useMemo(() => {
+    return (
+      typeof auctionId === 'number' &&
+      typeof from === 'string' &&
```
josojo, line 140: From is never used, is it?

src/hooks/useClaimOrderCallback.ts
```diff
@@ -186,10 +195,45 @@ export const useClaimOrderCallback = (
...
+  const { account, library } = useActiveWeb3React()
+  const { auctionId, chainId } = auctionIdentifer
+
+  const easyAuctionInstance: Contract = getEasyAuctionContract(chainId as ChainId, library, account)
+  const {
+    loading: isLoadingClaimed,
+    result: not_claimed,
+  } = useSingleCallResult(easyAuctionInstance, 'containsOrder', [
+    auctionId,
+    claimableOrders == undefined || claimableOrders[0] == undefined
```
josojo, line 233: if (claimableOrders == undefined || claimableOrders[0] == undefined) is true, then we don't need to make a call to the blockchain at all. Its sure that this will always return false

src/hooks/useClaimOrderCallback.ts
```diff
@@ -186,10 +195,45 @@ export const useClaimOrderCallback = (
...
-      summary: 'Claiming tokens',
+      summary: `Claiming tokens auction-${auctionId}`,
     })
     return response.hash
   }, [account, addTransaction, chainId, error, gasPrice, library, auctionId, claimInfo])
 
-  return claimCallback
+  const claimableOrders = claimInfo?.sellOrdersFormUser
+  const userHasAvailableClaim = useUserHasAvailableClaim(auctionIdentifier, claimableOrders)
+  const pendingApproval = useHasPendingClaim(auctionIdentifier.auctionId, account)
```
josojo, line 205: Here a renaming would be appropiate

josojo (commented):
Really good! Cool that you were able to come up with this solution so quickly!

Just left minor comments. I will test it later

src/state/transactions/hooks.tsx
henrypalacios: I forgot to put `from` in the statement, for the sake of context. But it would not be necessary since that `useAllTransactions` brings them filtered.

src/hooks/useClaimOrderCallback.ts
henrypalacios: I have renamed.

josojo:
>In Addition, I believe that the responsibility for consulting claimableOrders could be removed from useDerivedClaimInfo.

Yeah, I also dislike this duplication of code. 
useDerivedClaimInfo
and 
useUserHasAvailableClaim

are doing the same call. I think we should remove one or the other. 

I don't see how to derive correctly the error message, if claimableOrders is not doing the rpc call. Hence, I would try to merge these two functions or at least let both of them consume the same rpc-call.

src/hooks/useClaimOrderCallback.ts
josojo: I see. I am also not sure how to do it better.

josojo (approved):
I would really like to release this already. It's such a pain for users. 

Let's tackle the duplication in code / in calls in another PR. @henrypalacios please create an issue continue working on this.

Issue #577, opened 2021-05-19
Avoid 2 RPC calls about claim info from Claimer component


henrypalacios:
After adding the Claim status tracking [PR-571](https://github.com/gnosis/ido-ux/pull/571), a duplicate call has been added to [`useDerivedClaimInfo`](https://github.com/gnosis/ido-ux/blob/656386fec87a17a1a6223b3b9bfe0be61f0148cc/src/state/orderPlacement/hooks.ts#L546-L561)

PR #592, opened 2021-05-20
Add useGetClaimState

henrypalacios:
closes: #577 

Avoid another RPC call about from claim info, used in `useDerivedClaimInfo`

**Enviroment**:
auctionId: 90
chainId: 4

### DIFF
--- a/src/hooks/useClaimOrderCallback.ts
+++ b/src/hooks/useClaimOrderCallback.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 import { TokenAmount } from 'uniswap-xdai-sdk'
 
 import { BigNumber } from '@ethersproject/bignumber'
@@ -218,7 +218,6 @@ export function useGetClaimState(
   const [claimStatus, setClaimStatus] = useState<ClaimState>(ClaimState.UNKNOWN)
   const { account, library } = useActiveWeb3React()
   const { auctionId, chainId } = auctionIdentifier
-  const accountRef = useRef(account)
 
   useEffect(() => {
     setClaimStatus(ClaimState.UNKNOWN)
@@ -237,12 +236,12 @@ export function useGetClaimState(
 
     async function userHasAvailableClaim() {
       try {
-        if (!library || !accountRef.current || !claimableOrders) return
+        if (!library || !account || !claimableOrders) return
 
         const easyAuctionContract: Contract = getEasyAuctionContract(
           chainId as ChainId,
           library,
-          accountRef.current,
+          account,
         )
 
         const method: Function = easyAuctionContract.containsOrder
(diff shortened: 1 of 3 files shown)


src/hooks/useClaimOrderCallback.ts
```diff
@@ -201,39 +204,72 @@ export const useClaimOrderCallback = (
...
-  } = useSingleCallResult(easyAuctionInstance, 'containsOrder', [
-    auctionId,
-    claimableOrders == undefined || claimableOrders[0] == undefined
-      ? queueStartElement
-      : claimableOrders[0],
-  ])
-
-  return Boolean(!isLoadingClaimed && not_claimed && not_claimed[0] === true)
+  const { auctionId, chainId } = auctionIdentifier
+  const accountRef = useRef(account)
```
josojo, line 221: Why are you using a ref here?

src/hooks/useClaimOrderCallback.ts
```diff
@@ -1,19 +1,21 @@
...
 import { AuctionIdentifier } from '../state/orderPlacement/reducer'
 import { useHasPendingClaim, useTransactionAdder } from '../state/transactions/hooks'
 import { ChainId, calculateGasMargin, getEasyAuctionContract } from '../utils'
+import { getLogger } from '../utils/logger'
 import { additionalServiceApi } from './../api'
 import { decodeOrder } from './Order'
 import { useActiveWeb3React } from './index'
 import { useGasPrice } from './useGasPrice'
 
+const logger = getLogger('useClaimOrderCallback')
```
josojo, line 17: Is it still needed?

josojo (commented):
Cool. Looks good!

src/hooks/useClaimOrderCallback.ts
henrypalacios: to avoid unnecessary calls to `userHasAvailableClaim` when changing the `account` assignment, and make the call only when the value of the [`account` changes](https://github.com/gnosis/ido-ux/pull/592/files#diff-99c45e8374392fb1f48169b1d9b6ba52c3e15e2666b8b4e1ab940ab5ca1a2ac1R272).

src/hooks/useClaimOrderCallback.ts
henrypalacios: I am not sure if we have an error handling policy but I thought it might be useful to register the [unavailability of EasyAuction services](https://github.com/gnosis/ido-ux/pull/592/files#diff-99c45e8374392fb1f48169b1d9b6ba52c3e15e2666b8b4e1ab940ab5ca1a2ac1R264).

if you find it unnecessary, please let me know.

src/hooks/useClaimOrderCallback.ts
ramirotw: isn't this going to prevent re-running the hook when the user changes their wallet's address?

henrypalacios: no, because the change in account will restart the status cycle in `setClaimStatus(ClaimState.UNKNOWN)`, what will happen is that `userHasAvailableClaim` will query again with the address that was before the change 😥. 

I will fix it, thanks for the questioning @ramirotw

Issue #344, opened 2021-03-29
Update auction info after order placement or cancellation


elena-zh:
When I place an order or cancel an order, I do not see an updated info on the auction details page. E.g. current price, chart, sometimes even details section are not updated and show the previous info until I manually reload the page. 
https://drive.google.com/file/d/1ohO3jhS47Xt1Ouu7WKSVZZVngbTdBkX-/view

It would be great to auto-update the page after receiving a 'successful' order placement/cancellation transaction response.

josojo:
@henrypalacios 
for tthe orderbookchart, we already update the clearing price - the white line. You can just use this logic to calculate the new clearing price and display itt.

henrypalacios:
@josojo I would like to know where the `current price` that is on the chart, I have hunch it is from [here](https://github.com/gnosis/ido-ux/blob/d712b0859991935906105f4b77e9a14df80bd37a/src/components/auction/OrderbookWidget/index.tsx#L332). But I still can't trace the output of the `currentPrice` value in  `proccessOrderbookData` function.

elena-zh:
Hey @henrypalacios , 
I'm not sure if you have taken into account these cases, that is why I would like to mention them here. 
We have additional details section in an auction details page that shows 'Minimum Funding' and 'Estimated Tokens Sold', and it would be nice to recalculate these numbers as well when a user places/cancels orders.
![image](https://user-images.githubusercontent.com/70885163/121196097-dc256100-c878-11eb-937c-87e05247d1a7.png)

Could you please check it?
Thanks

cmagan:
I'm not sure what input you need from me on this issue...
I think it'd be very useful to update the "current price", "minimum funding" and "estimated tokens sold"
The thing is, we should update them only once the order transaction has been included in a block

josojo:
>The thing is, we should update them only once the order transaction has been included in a block

why do you think we should do it that way? isn't is clearer to the user, if there is only on coherent price shown in the whole app? In the past, people said that it's confusing that we show a different current price in the graph compared to the auction details.

I am quite unsure now...

cmagan:
Yeah we should def update both at the same time!
I was just saying that it should be updated after the tx is confirmed and not just after submitted...

josojo:
>I was just saying that it should be updated after the tx is confirmed and not just after submitted...

but right now, it is updated just on order placement indication, since the new order is getting into the graph

cmagan:
I think it's two different features.
[...]

PR #629, opened 2021-05-29
Update clearingPriceInfo after blockNumber change

henrypalacios:
closes: #344 

The problem is that the [Hook](https://github.com/gnosis/ido-ux/blob/f69471e685c86d84882cb542a13ee67dafd67e34/src/hooks/useCurrentClearingOrderAndVolumeCallback.ts#L10) in charge to update the `clearingPriceInfo` does not update at the moment to put new order or after block number change.

**Proposal:** Add `orderbookPrice: number` and `orderbookPriceReversed: number` fields to `orderbook` state reducer 
   

**Enviroment:**
auctionId: 114
chainId: 4


elena-zh:
The issue is still reproducible for me: prices (max bid and a current one) are not changed after placing an order;
chart, prices are not updated after cancelling an order.
https://drive.google.com/file/d/1t90Qwb8LJd3D7AUF2FeyLgehcg6afLjQ/view

src/state/auctionPrice/hooks.ts
```diff
@@ -0,0 +1,126 @@
...
+      } catch (error) {
+        onResetCurrentPrice()
+        console.error('Error getting clearing price info', error)
+      }
+    }
+
+    fetchApiData()
+    return (): void => {
+      cancelled = true
+    }
```
hpmaxi, line 117: @ramirotw What do you think is the best way to handle this pattern? Maybe checking `fetchApiData().then(() => { if(!cancelled...)` or something like that. Or a hook with support for async functions.

src/state/auctionPrice/hooks.ts
```diff
@@ -0,0 +1,126 @@
...
+export function useSetCurrentPrice(
+  auctionIdentifier: AuctionIdentifier,
+  derivedAuctionInfo: DerivedAuctionInfo,
+) {
+  const { shouldLoad: shouldLoadPrice } = useAuctionPriceState()
+  const { onResetCurrentPrice, onSetCurrentPrice } = useAuctionPriceHandlers()
+  const {
+    current: { auctionId, chainId },
+  } = useRef(auctionIdentifier)
+
```
hpmaxi, line 80: @henrypalacios I guess this is for performance, can you add a comment about this for future readers?

josojo:
Hey!
sorry for jumping in so late :(

>@josojo I would like to know where the current price that is on the chart, I have hunch it is from here. But I still can't trace the output of the currentPrice value in proccessOrderbookData function.

yes, this hunch was correct. You could use this method in the reducer to calculate the price.

I would highly prefer that, as the backend-api might not be updated. There are delays since FE and backend could be talking to different ethereum nodes and it might happen that the call to the API still returns the old price.

elena-zh:
Hi @henrypalacios , I have verified this issue. Great job! Current price/chart/orders sections are updated in a real time now. 
However, Minimum Funding/Estimated Tokens Sold and Max (Min) Bidding Price fields are not yet updated in a real time. I have created a separate issue for this case #744 , so we can start working on it in another PR. 

Thanks!

src/state/auctionPrice/hooks.ts
henrypalacios: You're right, it's not necessary for primitive values.

henrypalacios:
@josojo could you give me a hint on how to calculate the `priceReversed` with the [bids with format `PricePoint`](https://github.com/gnosis/ido-ux/blob/d712b0859991935906105f4b77e9a14df80bd37a/src/components/auction/OrderbookWidget/index.tsx#L288)

josojo:
The pricepoint data is never reverted

> @josojo could you give me a hint on how to calculate the `priceReversed` with the [bids with format `PricePoint`](https://github.com/gnosis/ido-ux/blob/d712b0859991935906105f4b77e9a14df80bd37a/src/components/auction/OrderbookWidget/index.tsx#L288)

I think I don't understand your question. Why doesn't
reversePrice( findClearingPrice(..)) work?

josojo:
You have written a whole new reducer logic for a new price.
It looks written well, though, we could keep the complexity much lower, if we just add a new field:
price to the state of the orderbook.
```
const {
    asks,
    auctionId: orderbookAuctionId,
    bids,
    chainId: orderbookChainId,
    error,
    userOrderPrice,
    userOrderVolume,
  } = useOrderbookState()
```
Then we also don't need to trigger more redux operations, as the newest order and the whole orderbook data is already maintained in the `orderbookstate`

Do you think that this would reduce the complexity quite a bit?

henrypalacios:
Thansk for the feedback @josojo, maybe I am overestimating the problem thinking about storing the [**AuctionDetails later**](https://github.com/gnosis/ido-ux/issues/344#issuecomment-856793158)

I agree the proposed change would reduce complexity, let me do it.

henrypalacios:
@josojo I found 2 approches with their pros and cons:
1. Calculate `clearingPrice` Locally after having the data from `onResetOrderbookData`  API. **(implemented)**
:heavy_check_mark: The `currentPrice` will be syncronized with the orderlist displayed in the UI.
:negative_squared_cross_mark: There may be cases where the node connected to the backend is not up to date.

2. The first approach + calculate locally after **place**/**cancel** order.  
:heavy_check_mark: The `currentPrice` will be immediately displayed in the UI regardless of the node's outdatedness.
:negative_squared_cross_mark: The `currentPrice` may be out of sync with the orderlist displayed in the UI.
:negative_squared_cross_mark: It Would require a **dispatch** to update the `currentPrice` only, to use only when changing the `auctionId` and not overwrite the one in the **state**.

josojo:
> Calculate clearingPrice Locally after having the data from onResetOrderbookData API. (implemented)
✔️ The currentPrice will be syncronized with the orderlist displayed in the UI.
❎ There may be cases where the node connected to the backend is not up to date.

I don't see the ❎: the orderbook also gets updated with the user order, isn't it?

josojo:
My approach would be the folloiwng:

- the shown price should be always the same as the current price shown in the chart: white dashed line
- the updating behaviour should be the same as on the chart.


The easiest to implement is probably:
- adding a new state to the orderbook state called: `orderbook price`.
- recalculating the orderbook price everytime the orderbook state is modified

And then we also don't need to pull in any new hooks etc...

henrypalacios:
Please, @josojo, @ramirotw  what do you think about the new changes?  

I took into consideration:
- [x] the updating behaviour should be the **same as on the chart**.
- [x] adding a new state to the `orderbook` state called: `orderbookPrice`.
- [x] recalculating the `orderbookPrice` everytime the `orderbook` state is modified.
- [x] don't pull any new hook, don't trigger any new redux operation.

src/state/orderbook/hooks.ts
```diff
@@ -56,9 +64,19 @@ export function useOrderbookActionHandlers(): {
...
+      const initialAuction = asks[0]
+      const userOrder = { price: Number(price), volume: userOrderVolume }
+      let calculatedAuctionPrice
+      if (initialAuction) {
+        calculatedAuctionPrice = CalculatorClearingPrice.fromOrderbook(
+          bids,
+          initialAuction,
+          userOrder,
+        )
+      }
```
josojo, line 76: Why are you doing the calculation in here and not in the `orderbook/reducer.ts`?
Wouldn't it be more performant to do it there?

josojo (commented):
Cool! Looks good! wiill reviiew it later more deeply

src/state/orderbook/hooks.ts
henrypalacios: I had doubts about where to place this, but this way my criteria:
1. I have seen how in most of the reducers in the project they are only used for state assignment.
2. Decouple the business logic of the library `redux`. (I still can't figure out how it could be more effective to do it there,  maybe I'm missing something)

But in case you suggest to move it to the `reducers`, I suppose that it would be in the cases of `resetUserVolume` and `resetOrderbookData` so that it continues to comply with:
>  the updating behaviour should be the same as on the chart.

elena-zh:
The issue is verified. 
Current price/chart/orders sections are updated in a real time. 
Please, be aware of the recent issue that was fixed and is already in dev, but still exists in this pr:
image

henrypalacios:
@elena-zh Could you tell me the issue you are referring to and how i can reproduce it.

elena-zh:
@henrypalacios , it was fixed in #766 and #768 prs

henrypalacios:
@elena-zh there was a mismatch with the merger, could you please try again?

elena-zh:
HI @henrypalacios , 
the issue is fixed. 
However, I'm not sure if we should change current price at the top of the page when an order has not been placed yet. Yes, the chart shows an approximate current price value, but the current price at the top should remain to show an actual one until we place an order, IMO.. @josojo , what do you think?
https://drive.google.com/file/d/11Cs2FZItCGKEjxgD_kvPqQwU9W_TiJzc/view

josojo:
I think it's less confusing for the users to see only 1 coherent "current price". Wdyt @cmagan ?

But yeah, would most amend the tooltip to state that the current user order is considered in the clearing price.

henrypalacios:
@cmagan I'd like to give you more context to make a decision regarding this PR ([PR-UI review](https://pr629--idoux.review.gnosisdev.com/)).

**This is the summary of this implementation:**
- The `current price` will be calculated and displayed at the time of typing in the **place order** form, (as it works in the `current price` **dot line** of the chart).

After reading the [comments](https://github.com/gnosis/ido-ux/issues/344#issuecomment-878349072) with @josojo, we must decide between these 2 approaches:

1. The updating `currentPrice` behavior should be the same as on the chart marked by **white dashed line**, (As implemented in this PR)

2. Change to comply:
> we should update them only once the order transaction has been included in a block

josojo:
I talked to Chen, and our conclusion is that indeed the price should only update, once the tx placement is finalized.

henrypalacios:
@elena-zh The suggested implementation is:
- [x]  update `currentPrice` only once the order transaction has been included in a block

I have restored some changes, but it should be fine.

elena-zh:
@henrypalacios , works great. Thank you!

Issue #646, opened 2021-06-03
'NAN' is displayed in the 'Min Bidding Price' for a newly started auction


elena-zh:
create an auction and open its details

**AR**: 'NAN' is displayed in the 'Min Bidding Price' for a newly started auction. Moreover, it is impossible to switch it to the Max price
https://drive.google.com/file/d/14-33K4jjp2tWEFgQOoNDpodtgy5jMJ3H/view
![nan.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/20e0a0ad-0844-41cb-8a1d-5143b482cbbf)

**ER:**  'Min Bidding Price' should show current price+1%.

auctionId=130&chainId=4
auctionId=129&chainId=4
auctionId=128&chainId=4

elena-zh:
Fixed in #665 
Case when it is impossible to switch prices is reported in a separate ticket #668

PR #665, opened 2021-06-09
Making getInverse aware of null state -

### LINKED ISSUE #616: [place order] wrong validation when set Max bidding price to 0
1. open app
2. open some active public auction
3. fill in the Amount field
4. Fill in the Max bidding price field with 0

**Actual Result:** validation message under the Amount field appears

**Exp.result:**  validation under the Max bidding price field appears: 'Max bidding price must be bigger than <>'

josojo:
closes #616

### DIFF
--- a/src/components/auction/Orderbook/index.tsx
+++ b/src/components/auction/Orderbook/index.tsx
@@ -59,8 +59,8 @@ export const OrderBook: React.FC<OrderbookProps> = (props) => {
 
   if (showChartsInverted(baseToken)) {
     for (const p of processedOrderbook) {
-      p.priceNumber = getInverse(p.price, NUMBER_OF_DIGITS_FOR_INVERSION)
-      p.priceFormatted = getInverse(p.price, NUMBER_OF_DIGITS_FOR_INVERSION).toString()
+      p.priceNumber = Number(getInverse(p.price.toString(), NUMBER_OF_DIGITS_FOR_INVERSION))
+      p.priceFormatted = getInverse(p.price.toString(), NUMBER_OF_DIGITS_FOR_INVERSION)
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/components/auction/Orderbook/index.tsx
```diff
@@ -59,8 +59,8 @@ export const OrderBook: React.FC<OrderbookProps> = (props) => {
 
   if (showChartsInverted(baseToken)) {
     for (const p of processedOrderbook) {
-      p.priceNumber = getInverse(p.price, NUMBER_OF_DIGITS_FOR_INVERSION)
-      p.priceFormatted = getInverse(p.price, NUMBER_OF_DIGITS_FOR_INVERSION).toString()
+      p.priceNumber = Number(getInverse(p.price.toString(), NUMBER_OF_DIGITS_FOR_INVERSION))
```
josojo, line 62: This is the only place where a problem would arise, as Number(`-`) = NAN

But since   p.priceNumber = NAN is not causing problems, AFAIK, I would leave it that way. I am not sure how to better deal with it. Not showing the orderbook in such a case is not an option for me.

elena-zh:
Hey @josojo , 

The linked #616 issue  seems to be still reproducible in this pr: 
image

However, another issue,  #646, is not reproducible there. 
Could you please clarify which bug you have fixed there: #646 or  #616?

Thanks

josojo:
>However, another issue, #646, is not reproducible there.
Could you please clarify which bug you have fixed there: #646 or #616?

These issues are very much related. Let me both tackle in this PR

josojo:
closes #646 as well.

henrypalacios (approved):
Great!

elena-zh:
@josojo , 
I'm still able to reproduce the #616 issue after the latest changes=(
image

Please, open this auction auctionId=138&chainId=4 and follow the next steps:
1. fill in an amount with a valid value
2. fill in the price field with 0
then, you will receive an irrelevant validation message for the Amount field, however, validation message for the price field should be displayed.

elena-zh:
@josojo , now everything LGTM!
Thank you!

Issue #660, opened 2021-06-09
Unwrap button is displayed all the time in an ended auction


elena-zh:
Unwrap button is displayed all the time in an ended auction even when:
1. User is not connected to a wallet
![not connected.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/61cbb508-0940-4491-be67-05a25028163b)
2. User is connected to another network
![another network.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/2e8d8d67-4fec-4b3a-94c9-bf951e601422)
3. User has 0 amount to claim (did not participate in an auction)
![0 amount.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/5bbc55e0-18b7-49d4-a67a-c4b95756cbe3)

**Expected**: Hide 'Unwrap button for the mentioned above cases

PR #669, opened 2021-06-10
Use native eth in order placement

### LINKED ISSUE #636: Enable place orders via ETH
Allow users to submit orders with ETH instead of WETH on mainnet and rinkeby

-> Show ETH balances in all interfaces instead of WETH
-> Make sure that the order placement uses the same logic as xdai orders and the right if-branch is used
https://github.com/gnosis/ido-ux/blob/d712b0859991935906105f4b77e9a14df80bd37a/src/hooks/usePlaceOrderCallback.ts#L110

-> Show wrap und unwarp buttons as currently implemented on xdai

hpmaxi:
Closes #636, #660 

Still need to test this. I guess `isTokenWETH` must be improved to works for rinkeby too. 

Also I think I should do something like `useTokenBalancesTreatWETHAsETHonXDAI`.

### DIFF
--- a/src/components/auction/Claimer/index.tsx
+++ b/src/components/auction/Claimer/index.tsx
@@ -12,7 +12,7 @@ import {
 import { useWalletModalToggle } from '../../../state/application/hooks'
 import { DerivedAuctionInfo, useDerivedClaimInfo } from '../../../state/orderPlacement/hooks'
 import { AuctionIdentifier } from '../../../state/orderPlacement/reducer'
-import { getTokenDisplay, isTokenXDAI } from '../../../utils'
+import { getTokenDisplay, isTokenWETH, isTokenXDAI } from '../../../utils'
 import { Button } from '../../buttons/Button'
 import { ButtonAnchor } from '../../buttons/ButtonAnchor'
 import { ButtonType } from '../../buttons/buttonStylingTypes'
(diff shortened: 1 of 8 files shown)


josojo:
Nice!
just added the small rest for testing. It works already great. I was able to place an order here:
`auction?auctionId=144&chainId=4#topAnchor`

The unwrap button can be linked to uniswap, instead of honeyswap on mainnet and rinkeby

elena-zh:
Although the pr is in draft, I have taken a look into it. 
Great job! I was able to place an order in ETH. 

However, here are my questions: 
1. When I cancel an order, I do not receive ETH back to my wallet. I get WETH. Is that OK? 
2. Unwrap button navigates to the https://app.honeyswap.org/ which is relevant for xDAI. I think, the link should be changed for Mainnet/Rinkeby networks. As long as tooltips for this button (tooltips are different for ongoing and ended auctions)
image
3. The issue #660 seems to be unfixed. I'm still able to see 'Unwrap' button for ended auctions where I did not participate/disconnected/connected to another network.

Thanks

josojo:
>When I cancel an order, I do not receive ETH back to my wallet. I get WETH. Is that OK?
yes, unfortunately, this is not possible with the current smart contracts

>Unwrap button navigates to the https://app.honeyswap.org/ which is relevant for xDAI. I think, the link should be changed for Mainnet/Rinkeby networks. As long as tooltips for this button (tooltips are different for ongoing and ended auctions)

Yes, unwrap buttons still need to be changed

elena-zh:
> yes, unfortunately, this is not possible with the current smart contracts

@josojo , what do you think about warning a user about this before cancellation? As an example, add a message to the cancellation confirmation pop-up or so.  I can create an enhancement for this case.

josojo:
> > yes, unfortunately, this is not possible with the current smart contracts
> 
> @josojo , what do you think about warning a user about this before cancellation? As an example, add a message to the cancellation confirmation pop-up or so. I can create an enhancement for this case.

good suggestion! let's not do it in this PR, but we can do it later with a new tickett

elena-zh:
Thanks, Alex! I have created #672 enhancement for this case.

hpmaxi:
@josojo @elena-zh Made some changes based on the feedback.

elena-zh:
@hpmaxi , 
Still see Honeyswap + xDAI tokens in tooltips. Unwrap button still redirects to the Honeyswap..

josojo:
Ups, I only fixed the text. I will also fix the link soon

josojo (approved):

elena-zh:
@josojo , @hpmaxi , 
changes look good to me! Great job, thank you!

PR #687, opened 2021-06-14
Feature/#533 highest volume auctions

### LINKED ISSUE #553: [UI] Showing highest volume auctions
In order to give new users a quick overview of how great auctions could look like, I think we should have a new section at the bottom of the landing page, that shows the highest volume auctions;

It could look like that:
https://docs.google.com/presentation/d/1RxkQqPepdSgr58wPXLscWd5BBmxen6uOs7AVwyOktcM/edit?usp=sharing

maria-vslvn:
Closes #553 

1. Highest volume auctions table at the mane page created
2. added mobile view
3. featured auctions data was used as mockdata


src/components/auctions/HighestVolumeAuctions/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+              </StyledTd>
+              <StyledTd>
+                <span>Buy Ammount:&nbsp;</span>
+                {auction.order.volume + ' ' + auction.symbolAuctioningToken}
+              </StyledTd>
+              <StyledTd>
+                <span>USD Volume:&nbsp;</span>
+                {'$' + auction.order.volume}
+              </StyledTd>
+              <StyledTd>{moment(new Date()).format('DD/MM/YYYY')}</StyledTd>
```
ramirotw, line 274: let's just use Date's `toLocaleDateString` for formatting and also remove momentjs library from the dependencies to reduce the bundle size

josojo:
Cool, looks already good.

Here are some suggestions:
Could we get the same column highlighting, as we have currently on the auction overview table?
Could we get the whole column clickable, and not just the > symbol?
Ammount -> Amount
Improve the mobile mode to something that looks better?
image
Maybe @alongoni can help here too

elena-zh:
Hi @maria-vslvn , @josojo , 
I have reviewed the changes, and here are mu notes: 
1. It would be nice to improve the landing page layout as currently we have different paddings between sections: 
different paddings
So, it would be good to have the same one between all the sections.

2. All the lines in the table are sorted in the BE by usdAmountTraded desc. However, sorting seems to be incorrect. See on the 4 and 5 lines: for  4 line tje usdAmountTraded = 5, for 5th line  usdAmountTraded=9. 
usdamounttraded

3. From my perspective, 'End Date' is not clear to be when see an auction details in the mobile view. I see a date, but I, as an end user, have no clue what date it is: ending/auction creation/etc. Could we add a label for this field?
image

4. Cards in the 'Featured Auctions' and 'Highest Volume Auctions' have different width in mobile/tablet view on a certain screen resolutions. I think, cards should look the same
the same width

5. This case might be fixed by the previous one, however, I will mention it. We have 5 auctions in the table, so on a certain screen resolutions we may get a bit not pretty view with an empty space from the right. I think, it would be nice to show whether 4 or 6 auction in order to avoid this case:
4 or 6

6. The last case.. We do not have a task for a logic implementation for this table. Currently, 'Featured auctions' are displayed in the table..

Thanks

alongoni:
Hey @maria-vslvn @josojo 
Regarding point 1) I suggest use this margin/paddings
02- GnosisAuction - Landing v2


For points 3) and 4) Here are some options to show the info in cards.
image

maria-vslvn:
1. same as "All auctions" row highlghting.
2. the  whole column was made clickable.
3. ammount -> amount.
4. moment lib was removed.
5. new indents were made between the landing sections.
6. all cards were made of the same width  on mob view.
7. new styles for highest vol auctions cards were applied for mob view, i applied the second variant.

josojo:
> Hey @maria-vslvn @josojo
> Regarding point 1) I suggest using this margin/paddings
> 02- GnosisAuction - Landing v2
> 
> For points 3) and 4) Here are some options to show the info in cards.
> image

Yes I like the margins a lot.
I like the third option card the best

elena-zh:
Hey @maria-vslvn , @alongoni , @josojo , 
I have verified the issue. Great job! 
The only thing I have to mention in addition to Alex comment about the 3rd option, is that it would be nice to center-align fields in the auction card as it is done for a featured auction card 
image

Cases 2 and 5 from my previous comment I have moved to #690 task. 

Thanks

maria-vslvn:
@josojo @elena-zh i've updated the auction card to 4variant. according to centering the data, i've spent some time on it, and it takes time to implement and show the visual center. if we really need this i can spent some more time

elena-zh:
@maria-vslvn , I have created a separate issue for fields alignment inside an auction card #695. 
Card's layout is changed and LGTM. 
Thanks!

alongoni (changes requested):
Looks really good! I made two little suggestions. Thanks.

PR #689, opened 2021-06-14
Avoid running effect again when price is changed manually

### LINKED ISSUE #668: Can no longer invert prices, if auction token is sable token
to reproduce, 
- connect to rinkeby
- go https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=143&chainId=4#topAnchor ( any token with a stable token as auctioning token)
- try to click invertPrice button

Expected behaviour: prices invertt

Observed behaviour: prices invert and immediately invert back

hpmaxi:
Closes #668

### DIFF
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -265,7 +265,8 @@ const AuctionDetails = (props: Props) => {
         onInvertPrices()
       }
     }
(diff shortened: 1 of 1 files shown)


josojo (approved):
A little bit hacky, but okay for me. It's better solution as introducing a more complex state,

elena-zh:
LGTM!

Issue #593, opened 2021-05-20
[auction details] fix behavior when enter more than 18 digitals after dot


tatyanakurinskaya:
https://images.zenhubusercontent.com/326940559/3ac72b06-cf56-4a93-b1fc-49da13508942/screen_recording_2021_05_20_at_09_10_28.mov

https://images.zenhubusercontent.com/326940559/5cd8448f-d781-4598-b01e-af787ca660fa/screen_recording_2021_05_20_at_09_12_50.mov

https://user-images.githubusercontent.com/863519/118932899-3ce40c80-b951-11eb-99bb-59ce02dd6e1c.mov

Issue #262, opened 2021-03-22
High price orders are not working


cmagan:
When trying to submit order with high price, the spinning wheel in the confirmation modal keeps spinning but no metamask pop up appears.
![image.png](https://images.zenhubusercontent.com/60227f03bd43d8095c1185d0/ade4c639-f479-44f4-8395-01c8aee1ccb6)

mariano-aguero:
The _buyAmountScaled_ calculated in the method _convertPriceIntoBuyAndSellAmount_ was zero, so the contract will trigger a revert exception. We are using the BigNumber library from ethers, maybe we need to use another library to calculate with more precision the value to use (for example Big.js)

josojo:
>The buyAmountScaled calculated in the method convertPriceIntoBuyAndSellAmount was zero, so the contract will trigger a revert exception. We are using the BigNumber library from ethers, maybe we need to use another library to calculate with more precision the value to use (for example Big.js)

Yeah, i think the restriction is within the smart contract. It can't deal with these numbers. I think a simple warning in UI that the price is too high and a user should choose a lower price is sufficient.

Sorry that I have not seen this earlier.

elena-zh:
Issues reported in this comment https://github.com/gnosis/ido-ux/issues/262#issuecomment-804728549 are fixed. 
The main issue in the description is still reproducible.

Issue #231, opened 2021-03-19
Incorrect validation messages when enter too long values into the price field


josojo:
If a user puts in a price as long as 4.000000000005, then the error message is wrong.
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/9d98c50f-9dbf-4e9d-996c-51355c55cf08)

Sometimes, the same validatiton message may appear when amount corresponds to the condition (I mean, when amount>required one)
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/cebf5cf0-b0b9-42c9-9038-d5eabad05384)

elena-zh:
Fixed in #691

PR #691, opened 2021-06-15
Not allow invalid amounts

### LINKED ISSUE #580: [auction details] place order button becomes enabled if amount=0
1. open app
2. follow Auctions from menu
3. open auction 
4. click MAX or enter 0
5. fill in the price with 0 as well

**Result:** Place order button becomes available 
 
**Exp.result:**  Place order button is disabled. 
Also, it would be nice to show validation messages for the fields when they are empty/filled in with 0 amounts

henrypalacios:
closes #580 

When an amount is placed, [a parse attempt is made](https://github.com/gnosis/ido-ux/blob/8accd6ba7b49d5a21e06859c0a5672da3890202c/src/state/orderPlacement/hooks.ts#L171-L186) on the bidding token.

**Proposal**: Invalidate amount when it could not be parsed.

- This will invalidate the 0.
- This would invalidate amounts exceeding the bidding token's decimal places. (I don't know why anyone would want to do that) related #593 

> :bulb: I don't think it is necessary but the error in case of decimals could be better specified.


hpmaxi (approved):

elena-zh:
Hi @henrypalacios , could you please provide a review link to test these changes?
Thanks

elena-zh:
@henrypalacios , #580 issue LGTM! However, I'd rather rewrite validation message for the Price field in a manner 'Price must me higher than 0', as such messages are shown when validation price on a min value.
higher image

Case #231 (#593 duplicates it) is not fixed in this PR: 
image

Thanks!

elena-zh:
Hey @henrypalacios , great! Now everything looks good to me (#580 and #231 cases).
The only 'curious' thing is that there is no validation for an inverted price field on number of decimals. 
However, I can successfully place an order with an inverted price with 20 decimals, and it works fine. Hence, there is no need to fix it and this issue can be closed. 

Btw, could you please take a look into the old #262 issue if you have a spare time? It is also related to the price field.

Thanks

src/state/orderPlacement/hooks.ts
```diff
@@ -216,32 +216,47 @@ export const useGetOrderPlacementError = (
...
   const insufficientBalance =
     balanceIn &&
     amountIn &&
     balanceIn.lessThan(amountIn) &&
     `Insufficient ${getTokenDisplay(amountIn.token, chainId)}` + ' balance.'
 
   const priceEqualsZero =
-    amountIn && price && (price === '0' || price === 'Infinity') ? `Price must not be 0` : undefined
+    amountIn && price && (price === '0' || price === 'Infinity' || price === '.' || price === '0.')
+      ? `Price must be higher than 0`
```
josojo, line 252: This is also not optimal, is it? 

Firstt the user gets the message: "higher than 0", and then next he gets the message "higher than x", where x is the current auction clearing price

henrypalacios: :bulb:  Let me unify the message.

elena-zh:
Hi @henrypalacios , changes for ordinary price LGTM!
However, 
1. when a price is inverted, it is a bit confusing to see than amount should be less than <min price> when we enter 0. 
image
It would be nice to save the logic for 'inverted prices' 
 - 0: 'Price must be higher than 0'
 - and then, if an amount is greater than min price, show 'Price must be lower than <min price>'

2. Then, the next case is related to an auction state when order cancellation is not allowed. Min price is used for validation
image


User should see message 'price should be higher than <current price>'  in such case

image

elena-zh:
@henrypalacios , I think it might help you

```
ordinary prices + orders are allowed to be cancelled
		price = 0 :  'price should be higher than <min bidding price>' 
		price < min sell price: price should be higher than <min sell price>' 
		
		
ordinary prices + orders are NOT allowed to be cancelled		
		price = 0 :  'price should be higher than <current price>' 
		price < min sell price: price should be higher than <current price>' 
		price > min sell price and < current price: price should be higher than <current price>'


inverted prices + orders are allowed to be cancelled
		price = 0 :  'price should be higher than 0' 
		price > max sell price: price should be lower than <max sell price>' 

inverted prices + orders are NOT allowed to be cancelled
		price = 0 :  'price should be higher than 0' 
		price > max sell price: price should be lower than <current price>' 	
		price < max sell price and > current price:	price should be lower than <current price>' 
```

src/state/orderPlacement/hooks.ts
```diff
@@ -220,28 +225,37 @@ export const useGetOrderPlacementError = (
...
+  const messageInitialPrice = () =>
+    showPricesInverted ? `Price must be higher than 0` : messageHigherClearingPrice
+
   const priceEqualsZero =
-    amountIn && price && (price === '0' || price === 'Infinity') ? `Price must not be 0` : undefined
+    amountIn && price && (price === '0' || price === 'Infinity' || price === '.' || price === '0.')
+      ? messageInitialPrice()
+      : undefined
+  const invalidSellAmount =
+    sellAmount && amountIn && price && !sellAmountScaled && `Invalid Bidding price`
```
josojo, line 258: nit: bidding price

josojo (approved):
LGTM

elena-zh:
Hi @henrypalacios , 
cool, thank you! 
I have found the last case that remains to be unfixed: 
```
ordinary prices + orders are allowed to be cancelled
		price = 0 :  'price should be higher than <min bidding price>'
```

Currently, it shows validation message 'price should be higher than <current price>' instead of the <min bidding price>
https://drive.google.com/file/d/1xSB7isUV8l2d0P1eDBuB3Q65scpiX8F7/view
Could you please fix it?

All the rest cases look good to me now. 

Thanks

elena-zh:
@henrypalacios , cool! Now all changes look good to me!

PR #697, opened 2021-06-16
Use decimal.js to ease calc of significant digits

### LINKED ISSUE #692: Better number display for vita auction
Hey guys,

I just tested how the numbers are displayed in the vita auction in the graph. (its only a. test auction, but with the right numbers)

Can you please make sure that in the green box more digits are shown for the price, such that people can better see their bidding amounts? Thx

hpmaxi:
Closes #692 

Maybe decimal.js is an overkill, but could be of good use to simplify other calculations without using `BigNumber.from(10).pow...`,


elena-zh:
@hpmaxi , @josojo ,
I have verified the issue: now numbers look good and the whole specified price/volume is displayed in the green box. 
However, sometimes a lot of decimals could be displayed. Is that OK?
a lot

Another case, that there is no need in displaying prices/volumes in K/M now as they are not rounded to 2 decimals now
2 511 k
If it is worth fixing, I can create a separate issue for this case as it is not a high priority one and does not affect an upcoming auction. 

Thanks

elena-zh:
I have created a separate issue #699 for the cases above

josojo (approved):

PR #713, opened 2021-06-21
fixing price display in feature auction card

### LINKED ISSUE #712: Add decimals to a current price on an auction card
When price is too small, 0 is displayed on an auction card. 
Need to add decimals to a current price on an auction card in  order to show it correctly.

josojo:
closes #712


elena-zh:
Great!

Issue #721, opened 2021-06-21
Granularity feature shows incorrect granularities


josojo:
The current vita-auction is no longer showing the right options of granularity of 0.0..1, but just 1000, 100, ...

elena-zh:
@josojo , I think the root cause of this issue is calculation of the default value in the granularity dropdown: it is calculated as  (MaxPrice-MinPrice)/10. Since we have an order with 1000 price, 100 becomes a default value in the list, and 2 ranges up and down are added there
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/fc731768-10ac-4e99-ba09-608e467b8e26)

So, what is the best option for you to fix this issue
 - add more ranges up and down for the default value
- recalculate somehow a default granularity value?

Thanks

Issue #704, opened 2021-06-18
Granularity dropdown can appear/disappear in the Orderbook table


elena-zh:
Related to #674

1. Open any auction
2. reload a page several times OR switch between several auction

**AR**: Granularity dropdown can appear/disappear
See the video https://drive.google.com/file/d/19o7NHqFllPjNBuuJS7NGgQX0RJ4cadiq/view

**ER**: Granularity dropdown is always displayed.

ramirotw:
This behavior is intended. The granularity dropdown is built based on current bids. If there are no bids or just 1 the dropdown doesn't make any sense. This is more a UX issue, should we always display it?

elena-zh:
Thanks, @ramirotw , I will change the issue priority to Low, as there is another case that granularity dropdown from a previously opened auction appears on the next auction details, but then disappears after reloading the page (in case if it not needed for an auction). And vice versa.

elena-zh:
This issue might be caused by a #722 hotfix, however, currently, when I place a 1st order, I get a 0 price in the Orderbook and I have no possibility to change granularity to see an actual price (especially in case if it is less than 1)
![image](https://user-images.githubusercontent.com/70885163/122892772-5ad8ce80-d34e-11eb-9c02-0433d0f59f8d.png)

Issue #690, opened 2021-06-15
Showing highest volume auctions - logic


elena-zh:
Related to #533 and #687

Implement logic for showing highest volume auctions:
1. **Ended** auctions (incl. private ones) with highest volume in USD should be displayed in the table
2. Table is sorted by USD amount desc
3. Take a look into case 2 in the comment https://github.com/gnosis/ido-ux/issues/687#issuecomment-861421144
![image](https://user-images.githubusercontent.com/70885163/122212729-d9e28880-ceb0-11eb-95e1-407e4f92b811.png)

4. Need to decide how many grid lines to display in the table (3/4/5/6?). Currently, we have 5 auctions in the table, so on a certain screen resolutions we may get a bit not pretty view with an empty space from the right. I think, it would be nice to show whether 4 or 6 auction in order to avoid this case:
![4 or 6](https://user-images.githubusercontent.com/70885163/122044729-83f4de80-cde5-11eb-829f-643091c6f1d5.jpg)

elena-zh:
@ramirotw , great job!

PR #722, opened 2021-06-22
better granularity

josojo:


josojo:
I am merging it already. tested it sufficiently for a hot-fix.

Works good enough to 

closes #721 and (from my point of view) #690

elena-zh:
Changes LGTM!
However, I have found one case related to the 1st order in an auction: automatically selected granularity does not reflects an actual price and a user is not able to change it. I have reported it in the existing issue https://github.com/gnosis/ido-ux/issues/704#issuecomment-865740091

Thanks

PR #650, opened 2021-06-03
Activate wallet-connect deeplink with qr-mode

henrypalacios:
closes: #562 

Activate walletconnect deeplink.

:warning: **Problems related to this PR:** 
- Functionality Issue: #671
- UI issues: #670, #659

PR #730, opened 2021-06-23
Handling app crashing when walletconnect is using another supported network

### LINKED ISSUE #658: WalletConnect: the app is crashed when connect a wallet with another supported network on auction details
1. Open any ongoing public auction details (as an example, in Rinkeby)
2. Connect a wallet in, as an example, Mainnet using WalletConnect

**AR**: the app is crashed. It is impossible to reconnect a wallet once again as it appears that the session is already established in the connected app. That is why I always get a crash when I try to connect a wallet again using walletconnect option

https://drive.google.com/file/d/1v3anF2Ppz9ait75Pi2JdGOhLQSu_aU9P/view

**ER**: the wallet is connected, message to switch a network is displayed

henrypalacios:
closes: #658 

When I connect via `walletconnect` from a network not supported by the specified the app is crash.

- [x] Fixed the app crashed when the user connect via `walletconnet` through **from the Auctions** `/overview` page, where it is possible to connect from different supported networks.

### DIFF
--- a/src/components/modals/WalletModal/index.tsx
+++ b/src/components/modals/WalletModal/index.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react'
 import { NavLink } from 'react-router-dom'
 import styled from 'styled-components'
 
-import { URI_AVAILABLE } from '@anxolin/walletconnect-connector'
+import { URI_AVAILABLE, WalletConnectConnector } from '@anxolin/walletconnect-connector'
 import { AbstractConnector } from '@web3-react/abstract-connector'
 import { UnsupportedChainIdError, useWeb3React } from '@web3-react/core'
 import ReactGA from 'react-ga'
@@ -188,7 +188,11 @@ const WalletModal: React.FC = () => {
       }
     } catch (error) {
       if (error instanceof UnsupportedChainIdError) {
-        activate(connector) // a little janky...can't use setError because the connector isn't set
+        let c = connector
+        if (c[chainId] instanceof WalletConnectConnector) {
+          c = c[chainId]
+        }
+        activate(c) // a little janky...can't use setError because the connector isn't set
       } else {
         setPendingError(true)
       }
--- a/src/components/modals/WalletModal/index.tsx
+++ b/src/components/modals/WalletModal/index.tsx
@@ -188,11 +188,10 @@ const WalletModal: React.FC = () => {
       }
     } catch (error) {
       if (error instanceof UnsupportedChainIdError) {
-        let c = connector
-        if (c[chainId] instanceof WalletConnectConnector) {
-          c = c[chainId]
-        }
-        activate(c) // a little janky...can't use setError because the connector isn't set
+        // a little janky...can't use setError because the connector isn't set
+        activate(
+          connector[chainId] instanceof WalletConnectConnector ? connector[chainId] : connector,
+        )
       } else {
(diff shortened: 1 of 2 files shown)


henrypalacios:
@ramirotw we can continue from here with @elena-zh [notes](https://github.com/gnosis/ido-ux/pull/650#issuecomment-866755834).

> Hi @henrypalacios , thank you for the fix!
> Now the app is not crashed, however, works a bit strange. Look:
> 
> 1. When I connect to another supported network, I receive an error message to change a network.
> 2. A lot of errors in console appear right after connection
>    here is the video: https://drive.google.com/file/d/10AGylZilk3K166nEHCnK2JbrZl_ae57r/view
> 
> It would be great to use the same flow as it works for Metamask connection:
> 
> * show a connected state with a message to select another network instead of the 'Connect wallet'  button when a user is connected to a supported network
>   image
> * show an error and do not connect when a user is not in supported network.
>   Metamask flow: https://drive.google.com/file/d/1eORx4AoXUnaavA-GPcuEIFyQaIN0kYY_/view

src/components/modals/WalletModal/index.tsx
```diff
@@ -188,7 +188,11 @@ const WalletModal: React.FC = () => {
       }
     } catch (error) {
       if (error instanceof UnsupportedChainIdError) {
-        activate(connector) // a little janky...can't use setError because the connector isn't set
+        let c = connector
```
josojo, line 191: I think that each variable name should always be written out. It just improves code readability a lot.

Maybe 
```
activate(connector[chainId] instanceof WalletConnectConnector ?connector[chainId], connector)

```
is a better syntax

henrypalacios: I agree with your suggestion. Thank you for it.

henrypalacios:
I found that I have to handle 2 cases.
1. When the user logs in **from the home** where it is possible to connect from different supported networks.
2. When the user connects from the **Auction view** where the `walletconnect` construct will only allow one network, the `auctionId`.

henrypalacios:
@josojo I believe that what @elena-zh  suggest:
> It would be great to use the same flow as it works for Metamask connection:

could be achieved if we [change](https://github.com/gnosis/ido-ux/blob/issue-658/src/connectors/index.ts#L51) the default `walletconnect` to a single one, 
```typescript
export const walletconnect = new WalletConnectConnector({
    rpc: { 4: NETWORK_URL_RINKEBY, 100: NETWORK_URL_XDAI, 1: NETWORK_URL_MAINNET },
    bridge: 'https://safe-walletconnect.gnosis.io',
    qrcode: true,
    pollingInterval: POLLING_INTERVAL,
  })
```
Is there any reason why we should keep it as an object?

henrypalacios:
Sumary:
 - Avoided app crashing when connecting through a network other than the one defined in the auction.
 - Avoided unmanageable walletconnect errors (console errors), by creating a local status [`walletConnectChainError`](https://github.com/gnosis/ido-ux/blob/08085c720ec786dc9da6fc86f9598c7daeac35c0/src/components/modals/WalletModal/index.tsx#L92)

elena-zh:
Hi @henrypalacios , good job!
The app is not crashed now.. And works pretty good. Yea, it does not 100% follow MM flow, but anyways, works much-much better. I faced only known issues (all of them are reported under the #736 epic). 

Thanks

PR #737, opened 2021-06-25
Highest volume auctions integration

### LINKED ISSUE #690: Showing highest volume auctions - logic

ramirotw:
Closes #690

### DIFF
--- a/src/api/AdditionalServicesApi.ts
+++ b/src/api/AdditionalServicesApi.ts
@@ -17,7 +17,8 @@ export interface AdditionalServicesApi {
   getAllUserOrdersUrl(params: UserOrderParams): string
   getAllUserOrders(params: UserOrderParams): Promise<string[]>
   getMostInterestingAuctionDetailsUrl(params: InterestingAuctionParams): string
-  getMostInterestingAuctionDetails(): Promise<AuctionInfo[]>
+  getMostInterestingClosedAuctionDetailsUrl(params: InterestingAuctionParams): string
+  getMostInterestingAuctionDetails(closedAuctions?: boolean): Promise<AuctionInfo[]>
   getAllAuctionDetailsUrl(networkId: number): string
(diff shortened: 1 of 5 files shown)


hpmaxi (approved):
LGTM!
Maybe I would suggest moving `4`, the number of auctions to show, to a constant.

src/api/AdditionalServicesApi.ts
```diff
@@ -296,11 +306,16 @@ export class AdditionalServicesApiImpl implements AdditionalServicesApi {
...
+    closedAuctions = false,
+  ): Promise<Maybe<AuctionInfo[]>> {
     try {
       const promises: Promise<Response>[] = []
       for (const networkId in this.urlsByNetwork) {
-        const url = await this.getMostInterestingAuctionDetailsUrl({
+        const fn = closedAuctions
+          ? this.getMostInterestingClosedAuctionDetailsUrl
+          : this.getMostInterestingAuctionDetailsUrl
+        const url = fn.bind(this)({
```
henrypalacios, line 318: I liked, what happened to the `await` sentence?

henrypalacios (approved):

src/api/AdditionalServicesApi.ts
ramirotw: the functions weren't async so it wasn't necessary

PR #738, opened 2021-06-25
Check min funding criteria on claimer

### LINKED ISSUE #694: Incorrect amount to claim tokens in case if Minimum Funding is not met
1. Create an auction with a Minimum Funding amount
2. Place several orders with a total amount< Minimum Funding
3. Wait till the auction ends, settle it
4. Verify an amount to claim +run claim transaction

**AR**: sell tokens are displayed for claiming, however, a user gets buy tokens back as a Minimum Funding amount  is not  met
see the image

https://rinkeby.etherscan.io/tx/0x0389b85270159fa37337efa69c347b3dbc67b9c9d500c014c4159207a1233f1d

**ER**: buy token amounts should be displayed for claiming

hpmaxi:
Closes #694

~~This is failing for `auction?auctionId=170&chainId=4` but is working for others that @elena-zh made for testing. I still need to fix that error~~. Tried to improve readability of code, but still needs a little more of it. Conditions should be extracted to variables with meaningful names but I wasn't sure about which names, the same applies to the amounts added to the tokens.


src/hooks/useClaimOrderCallback.ts
```diff
@@ -101,59 +102,87 @@ export function useGetAuctionProceeds(
...
+    }
+
+    return {
+      claimableBiddingToken,
+      claimableAuctioningToken,
+    }
+  }
+
+  // Min funding is reached
+  // For each order from user add to claimable amounts (bidding or auctioning).
```
henrypalacios, line 164: I liked. Now it's easier to understand

src/hooks/useClaimOrderCallback.ts
```diff
@@ -101,59 +102,87 @@ export function useGetAuctionProceeds(
...
   for (const order of claimInfo.sellOrdersFormUser) {
     const decodedOrder = decodeOrder(order)
-    if (JSON.stringify(decodedOrder) == JSON.stringify(derivedAuctionInfo?.clearingPriceOrder)) {
+    if (JSON.stringify(decodedOrder) === JSON.stringify(clearingPriceOrder)) {
```
henrypalacios, line 167: if you could do the same in these 3 conditions when minimum is reached, it would be great.

henrypalacios (approved):

src/hooks/useClaimOrderCallback.ts
```diff
@@ -101,59 +102,87 @@ export function useGetAuctionProceeds(
...
+
+  const currentBiddingAmount = new TokenAmount(
+    biddingToken,
+    BigNumber.from(auctionDetails.currentBiddingAmount)
+      .mul(BigNumber.from(10).pow(biddingToken.decimals))
+      .toString(),
+  )
+
+  const minFundingReached = currentBiddingAmount.greaterThan(minFundingThresholdAmount)
+  const currentPrice = clearingPriceOrder.buyAmount.div(clearingPriceOrder.sellAmount)
```
josojo, line 147: nit: For easier readability, I would prefer to call it clearingPrice

josojo (commented):
Cool, looks good.

I know I have never done it, but it would be great to get some tests for such logic-heavy parts. Otherwise, we will rely too much on Elena

elena-zh:
Hi @hpmaxi , I have verified the issue. 
Token amounts are calculated correctly when min funding is not met. 
However, orders' amount calculations when min funding is met are broken now:
- a 'boarder value' issue: amounts to claim are recalculated incorrectly when min funding is 100%
image
Transaction returns correct amount tokens to a user's wallet https://rinkeby.etherscan.io/tx/0xa7cc8b6369ffde4a11516f3ee2842c2c65948a7aba8608ad05e452ff63ff31c0

 - then, min funding is met (101%), however, not 100% selling tokens are sold:
image
Transaction returns correct amounts: https://rinkeby.etherscan.io/tx/0x94abc9197f77a3adacec65ea169fcdc155d41f14925195c9bd83d24c2e07ce00

  - Amounts to return = 0 when minimum funding is 110%
image
However, user should be able to get a partial amount of a selling token 
image
 
 - Amounts to return = 0 when  when a user is 100% winner of an auction: 
image


Could you please fix it?

Thanks

hpmaxi:
@elena-zh Some errors should be fixed now. Still no tests for this.

elena-zh:
Hi @hpmaxi , 

cases when 
 - min funding = 100%, estimated tokens sold =100% is fixed
 - min funding = 101%, estimated tokens sold < 100% is fixed


But,

1. the case when min funding = 100%, estimated tokens sold <100% is not fixed: 
image

2. incorrect calculation when min funding > 100%, estimated tokens sold=100%
image


Could you please take a look into these issues?

Thanks

hpmaxi:
Now there is only one test based on an auction provided by Elena, I found it difficult to extract the test cases from `ido-contracts`  or to create my own so just used the auction that I knew was working fine. Doing that and extracting the function for testing, I encounter that a previous refactor in this PR made a bug at doing N.div(M) being `M >> N` with BigNumbers is 0 because no fractional values exist. This wasn't happening before because it was multiplied just before by the volume.

Hope this fixes the rounding issues and the last two cases @elena-zh .

Besides that, more cases should be added. Also, I've extracted `isMinFundingReached`.

elena-zh:
HI @hpmaxi , great job!
I have run a lot of tests with a lot of auction with different conditions, and everything works fine except one case:
incorrect calculation of tokens when bidding token is ETH 0xc778417E063141139Fce010982780140Aa0cD5Ab
I have got incorrect amounts when min funding was >=100%
eth
Transaction returned me a correct amount to my wallet (100 dai)
https://rinkeby.etherscan.io/tx/0xdac6243168fef0b0d05c4dd6e7bdac4b7f137798a1e8c528c6b4818a82707bf3


image

image

Could you please take a look at it? 

Thanks

hpmaxi:
@ramirotw The last commit adds a test similar to one of the cases with 100% funding, but is failing, it should help to debug the problem with auctions like the one in the screenshots I've sent by DM.

josojo:
@elena-zh could you please check one more time whether your catch persists. 

My assumption is that the amounts are shown correctly, but the funding percentage is still off.

elena-zh:
@josojo , sure, just a moment

elena-zh:
@josojo , no, the issue is still there
1. here is the auction auctionId=242&chainId=4, where min funding =100% and estimated tokens sold = 100%. However, amounts to claim are incorrect: 
image

I expect to see here these amounts: 
image
Claim transaction returns correct amounts https://rinkeby.etherscan.io/tx/0x01b3f459c42e000e4c5462799bf8647973377d1c00e2933cc25e9789309dc5d9

2. here is the auction auctionId=243&chainId=4, where min funding >100% and estimated tokens sold > 100%. However, amounts to claim are incorrect: 
image

And claim transaction returns incorrect amounts as well... https://rinkeby.etherscan.io/tx/0x2828414339ac94a00d2cc73e826a4dd5aa6b0bf3c3b60ed295d8067949efd9ea


Moreover, Min Funding and Estimated tokens sold values has stopped to be calculated and always show 0 now: 
image

And calculation of amounts to claim is broken for 'usual' token pairs again (auction?auctionId=244&chainId=4)...
image

Transaction also returns incorrect amounts. Hash: https://rinkeby.etherscan.io/tx/0xdd1e3e4b05b86752a2d82befd8f45541c73f93bb8eca283fcfa5d7048eee8f8a

ramirotw:
@josojo I'm currently working on this issue, there are several paths in the `getClaimableData` function that are not covered and also the current test when the min funding is met is also wrong. I'll cover as much as I can later in a new commit.

elena-zh:
My cases for tests: 
```
1. 	min funding <100%: returns full amount of a Bidding token
2. 	min funding <100%, estimated tokens sold = 100%: returns full amount of a Bidding token 
3. min funding =100%, estimated tokens sold = 100%: returns full amount of an Auctioning token 
4. min funding =100%, estimated tokens sold < 100%: returns partial amount of an Auctioning token according to the bidding token amount/closing price
5. min funding > 100%, estimated tokens sold = 100%: returns full amount of an Auctioning token + excess of a bidding token 
6. min funding > 100%, estimated tokens sold < 100%: returns partial amount of an Auctioning token according to the bidding token amount/closing price
```

Run this suit of tests for 'usual' tokens and  ETH 0xc778417E063141139Fce010982780140Aa0cD5Ab bidding token in an auction.

Run this suit of tests for multiple orders in an auction as well.

josojo:
>And claim transaction returns incorrect amounts as well... https://rinkeby.etherscan.io/tx/0x2828414339ac94a00d2cc73e826a4dd5aa6b0bf3c3b60ed295d8067949efd9ea

>Transaction also returns incorrect amounts. Hash: https://rinkeby.etherscan.io/tx/0xdd1e3e4b05b86752a2d82befd8f45541c73f93bb8eca283fcfa5d7048eee8f8a

@elena-zh the tx hash can never show incorrect amounts. We need to display in the front-end the logic of the smart contracts and hence the logic that will later be executed. If the tx amounts are incorrect, then you will have found an error in the smart contract, which would be a real bummer.

Regarding auction 244: The threshold was not met: Yes you placed a huge order of 11DAI, but only 0.1 DAI could be matched at this price and hence, only 0.1 would have been sold. In this case with 0.1 DAI being sold, the auction has not met the funding threshold.

Regarding auction 243: Same thing! You have to place the same order at a higher price to test it.

Regarding auctionId=242&chainId=4: Good catch. We need to check it one more time.

elena-zh:
@josojo , thanks a lot for the explanation!
Will wait the next fixes and will take into account your notes during testing the pr

ramirotw:
@elena-zh could you run another pass taking into consideration Alex's comments?

elena-zh:
@ramirotw , the issue related auctionId=242&chainId=4 where min funding =100% and estimated tokens sold = 100% is still there. 

The issue with 'usual' tokens is not reproducible (rechecked according Alex comments).

the issue with ETH bidding token when min funding is >100% (retested based on Alex notes) is still there
image
https://rinkeby.etherscan.io/tx/0x0038509463dc6ea94791ddeb2f8d94cad90d27f0ef7ce874cacbcef3553db495
`auctionId=247&chainId=4`

ramirotw:
Thanks Elena, I could reproduce it locally.

@josojo quick question: In auction 248, why is `currentBiddingAmount=0` ?

elena-zh:
@ramirotw , issue #745 exists for this case

ramirotw:
@elena-zh I see, so the issue with auction 242 is also related to #745.

@josojo should clarify if this is a expected behavior because both issues depend on the same piece of data from the API.

josojo:
>@josojo should clarify if this is a expected behavior because both issues depend on the same piece of data from the API.

I see! in this case, the API returns 0 bidding amounts. I will then resolve it in the backend and no front-end updates are required.

josojo:
@elena-zh
can you revisit your auction and see whether its fixed now? the API backend returns now 0.01 instead of 0.
https://ido-v1-api-rinkeby.dev.gnosisdev.com/api/v1/get_auction_with_details/242

elena-zh:
Hi @josojo , 
I cannot start testing as currently the app is crashed when I try to open an auction details page. 
Could you please take a look at this issue #760 ?

Thanks

elena-zh:
@josojo , I think that something is wrong with Min Funding and Estimated tokens sold calculation: 
image
Could you please take a look?
Thanks

elena-zh:
@ramirotw , @josojo , 
 1. Min Fund <100%, Min Fund <1 token, auctionId=255&chainId=4: **incorrect values**
image

Transaction returns correct amount: https://rinkeby.etherscan.io/tx/0x07c7a54a7fdc3a6498076b6cfaa9bb0158f16cd22fd1206d7167dbc37194e856

2. The same is with ETH token. But, as I understand, the issue is not related to a token pair, but to min funding amount < or > 1.
Anyways, auctionId=256&chainId=4: **incorrect values**
image
Transaction returns correct amount: https://rinkeby.etherscan.io/tx/0x1f541a6f64feee9ec916332783fc073ab2b43a65af576ee698cf37ebfa80d686

3. Min Fund <100%, Min Fund <1 token, auctionId=260&chainId=4: **incorrect values**
image
Transaction returns correct amounts: https://rinkeby.etherscan.io/tx/0x83a41348262bcd746d54e9ae31fb685df3d6fbcc95f05391c31b9175fdf1caf9


4.  Min Fund >100%, Min Fund <1 token, auctionId=257&chainId=4: **correct values**
image
https://rinkeby.etherscan.io/tx/0xc2234ded354375df2d95b6cc8b430eed97997427f3c8b1085bef3895721cb2ec

5. Min Fund =100%, Min Fund <>=1 token, auctionId=258&chainId=4: **correct values**
image
https://rinkeby.etherscan.io/tx/0xfdf67f48e1c4b020c35e976d9b5ec8b0c3d492040c7808658ccbd7c7bd1a20e0

So, cases 1-3 still need to fix. 

Thanks

ramirotw:
@elena-zh please re-test. With the latest changes in the API, the format of some values was changed and the minimum funding reach function was returning wrong values.

elena-zh:
Great!

PR #740, opened 2021-06-25
Prevent division by 0 when inverting prices

### LINKED ISSUE #706: Orderbook price = Infinity when invert prices
1. Open any ongoing auction with orders
2. Select a 'high' granularity in order to get '0' range of a price
3. Invert prices

**AR:** Price is 'Infinity'
https://drive.google.com/file/d/1r9M_-rmM3KthU4ynEaAtpGM1VovTBB96/view
infinity

**ER:** price should not be 'Infinity'

ramirotw:
Closes #706

### DIFF
--- a/src/utils/prices.ts
+++ b/src/utils/prices.ts
@@ -9,9 +9,16 @@ export function getInverse(price: string, nrDigits: number): string {
   if (price == '-') {
     return '-'
   }
+
+  // prevent division by 0
+  if (price === '0') {
+    return price
+  }
+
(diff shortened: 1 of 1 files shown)


src/utils/prices.ts
```diff
@@ -9,9 +9,16 @@ export function getInverse(price: string, nrDigits: number): string {
   if (price == '-') {
     return '-'
   }
+
+  // prevent division by 0
+  if (price === '0') {
+    return price
```
josojo, line 15: I think I would prefer to return the unset price of '-'.
But both is fine.

elena-zh:
LGTM!

PR #741, opened 2021-06-25
Fix orderbook table percentage bigger than 100

### LINKED ISSUE #707: My size can be >100% in the orderbook list
Related to #527, #674

Create several orders like on the pic

One price should be a border one (4.5 or so)

Then select a bigger granularity (1)

**AR**: My size  >100%
https://drive.google.com/file/d/1X3vaoLOV1DACqCmVkU2A0Kgmsghsooe2/view
I think that the 'boarder' price amount is added to the higher range price (46+5)/46 = 110.87%

**ER**: my size cannot be >100%

ramirotw:
Closes #707


josojo (approved):

elena-zh:
LGTM!

Issue #647, opened 2021-06-03
[Auction details] Zeros in placeholders seem to be truncated


elena-zh:
1. Open an ongoing public auction details
2. Take a look into the Place order section

**AR**: Zeros in placeholders seem to be truncated
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/235c934f-3c3c-49ff-a24f-1b4e55ecebae)

**ER**: zeros are not truncated.

PR #750, opened 2021-06-29
Feature/#716 truncated bidding price on mobile screens

### LINKED ISSUE #716: Android: the last digit in a max/min bidding price value is truncated when a value is long
See images:

**Expected**: value is not truncated.

maria-vslvn:
Closes #716

1. As the truncated value is an input value so i decided to highlight the input better


.github/workflows/deploy.yml
```diff
@@ -1,7 +1,11 @@
 name: Deploy
 
 # Run on every commit
-on: push
```
josojo, line 4: starting from the latest develop branch would prevent such issues

josojo (commented):
LGTM, though I can not test it right now.

elena-zh:
@maria-vslvn , could you please provide a PR preview link?

maria-vslvn:
@elena-zh

.github/workflows/deploy.yml
ramirotw: I'll paste what I wrote in the other PR:

@maria-vslvn when starting a new ticket in a new branch I'd suggest to do the following so you don't pick the wrong commits as with the `renaming deployments` which prevents Github Actions to run the deployment script:

```bash
git fetch
git checkout -b feature/#479-mob-menu-issue origin/develop
```

This will ensure to always fetch the latest changes from origin and then create the branch from the latest commit.

elena-zh:
HI @maria-vslvn , 
Changes look good to me, however, it seems that the arros icon is shifted now to the center between a new container and the fields label
image


Could you please fix it? 
Also, we have one more issue #647 with truncated placeholder zeros in these containers. 
Could you please also take a look at it? 

Thanks

elena-zh:
@alongoni , could you please take a look at these fields? 
Are they look OK for you? May be we can make boarder of one of containers transparent or you have another solution? 

Thanks

alongoni:
I'm not sure about highlight an input inside another one. Maybe we can try to make the input value border transparent.

maria-vslvn:
so @elena-zh @alongoni what decision should i apply? as earlier we had an issue of truncated strings on mobiles here. and i decided to highlight the inputs where we had these strings in order to make it clear that the string is not actually just truncated but it is editable and it's an input value. and as i understood now, i should remove these highlighting borders. so maybe you could offer something more suitable to solve the issue:)

elena-zh:
@maria-vslvn , the thing is that the 2nd container in the Amount field may bring some inconveniences in #758  task implementation. 
Also, 2 boarders, especially when they are in red, look a bit excessive, that is why there was a proposal tp make one of them transparent and to think how to make a better solution of this issue fix, taking into account #758 issue.

alongoni:
I think we should continue using the inputs as we usually do (option 1). But if we need to add more relevance to the fact that each input is still editable, we can think in option 2 by adding a dark background to the input. 
image

What do you think @maria-vslvn and @elena-zh ?

elena-zh:
@alongoni , both options are OK for me. 
However, I have a question for the 2nd option: how will long numbers look  in these small black inputs?
image
Thanks

alongoni:
Maybe we can set a 22 digits max length and then add an ellipsis in order to truncate big numbers.
image

elena-zh:
Great!

josojo (approved):

PR #753, opened 2021-06-30
Add new settlement state to auction details

### LINKED ISSUE #293: Auction claiming is written although claiming is not yet available
Once an auction ends, the interface shows "Auction Claiming"
But if the settlement tx wasn't yet submitted, then the "claim" button is not yet available.
This might create user confusion

We should consider additional middle state: "waiting for settlement"
Then users know the auction has ended but it's clear they cant claim yet
Once the settlement occurred we can display "Auction Claiming"

ramirotw:
Closes #293

### DIFF
--- a/src/components/auction/AuctionBody/index.tsx
+++ b/src/components/auction/AuctionBody/index.tsx
@@ -77,7 +77,8 @@ const AuctionBody = (props: AuctionBodyProps) => {
                 derivedAuctionInfo={derivedAuctionInfo}
               />
             )}
-            {auctionState === AuctionState.CLAIMING && (
+            {(auctionState === AuctionState.CLAIMING ||
+              auctionState === AuctionState.PRICE_SUBMISSION) && (
               <Claimer
                 auctionIdentifier={auctionIdentifier}
                 derivedAuctionInfo={derivedAuctionInfo}
--- a/src/components/auction/AuctionTimer/index.tsx
+++ b/src/components/auction/AuctionTimer/index.tsx
@@ -175,7 +175,7 @@ export const AuctionTimer = (props: AuctionTimerProps) => {
 
   const auctionStateTitle = React.useMemo(() => {
     if (auctionState === AuctionState.PRICE_SUBMISSION) {
-      return <TextBig>Auction Closed</TextBig>
+      return <TextBig>Waiting settlement</TextBig>
     } else if (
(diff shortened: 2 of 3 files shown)


ramirotw:
I did a small refactor to set the auction state `PRICE_SUBMISSION` value based on on-chain data when there's no clearing price order submitted.

josojo (approved):
Nice, seems like very good refactor, especially, if it saves API calls

hpmaxi (approved):
Great refactor and perf improvement.

elena-zh:
HI @ramirotw , 
a new status is added, however, status is not displayed (loading..), Claimig proceeds section is missing when: 
1. User is not connected to a wallet, auction is not settled
2. User is not connected to a wallet, auction is settled
3. User is in another supported network, auction is not settled/not settled
4. User is in not supported network, auction is not settled/not settled
mainnet
not supported
220 not connected

Could you please fix these cases?

Thanks

ramirotw:
@elena-zh I tracked down the issue and it's connected to https://github.com/gnosis/ido-ux/issues/401

When no account is active we deactivate the latest block number fetch, and if there's no block number we also don't execute any multicall contract request, hence we have no way to know if the auction is either in the claiming state or waiting for settlement as that state is only known based on on-chain data (the API doesn't return this info).

@josojo what do you suggest here?

josojo:
>When no account is active we deactivate the latest block number fetch, and if there's no block number we also don't execute any multicall contract request, hence we have no way to know if the auction is either in the claiming state or waiting for settlement as that state is only known based on on-chain data (the API doesn't return this info).


good observation. I think I will have to add it to the backend.

The now, I would suggest to show that the `claiming state`, as this is the majority of the time the "truth"

ramirotw:
@elena-zh please re-test. The Loading issue should be fixed for disconnected accounts. Now it defaults to CLAIMING even though the auction could be in the WAITING SETTLEMENT when the user is not connected to any wallet.

elena-zh:
@ramirotw , 
changes LGTM. 
User is able to see 'Waiting Settlement' status in cases when he/she is connected to the required network, auction is not settled. 
In all the rest cases I mentioned above, user sees 'Auction Claiming' status. 

Also, I have found a low priority UI issue and reported it in a separate ticket #761. 

Thanks.

PR #766, opened 2021-07-06
Updates to deal with the new backend logic

josojo:
Once the backend is updated to the latest code, this should fix the issues seen on ido-ux deployment.


josojo:
I am merging this quickly, as customers are also testing in the dev env

Issue #759, opened 2021-07-05
Create a Media Kit and add it to the docs as a new page


cmagan:
More info on here
https://gnosisinc.slack.com/archives/C01LJGKRJ0L/p1620737267005100

alongoni:
Hey @cmagan I've found this:
https://www.notion.so/Gnosis-Comms-e092358814da46f3833d352f76292c47

Gnosis Auction Media Kit.
https://www.notion.so/Gnosis-Auction-Media-Kit-c6c712821e6c4ad29a2e5736db5c1ff4

alongoni:
@cmagan I made this first draft in Figma, then we can export the logo and variants to import them in the GA Doc's section.
https://www.figma.com/proto/Y7MLpZujRAOFPVkplvA7kV/Gnosis-Auction?node-id=2%3A7435&viewport=75%2C379%2C0.34553584456443787&scaling=min-zoom&page-id=2%3A7434%3Ffuid%3D789860674698606135

I have a question here: Which is the normal logo and which is the variant? In the dApp we are using only the GA logo in white over a dark background and orange as an accent color. It's ok?

cmagan:
thanks @alongoni 
The media kit should have "Built using Gnosis Auction" as the main text
And in smaller text: "The fairest mechanism to auction assets on Ethereum" 
https://docs.google.com/presentation/d/1IuKY_NDEBemvcXB9275d6WfwW9i_59RU7FGyygyPe3g/edit#slide=id.g7a31c7822b_0_151

I think the official logo is the orange one, but it's fine to have also dark / light versions depending on the background

cmagan:
Would you be able to work on this today? @alongoni

PR #767, opened 2021-07-06
Media kit

alongoni:
Related to #759 
GA-dark-bg

The media kit was created in Docs section:
https://pr767--idoux.review.gnosisdev.com/#/docs/media-kit


elena-zh:
HI @alongoni , 
Could we add section names for the 1st 2 images? Something like 'Black background' and 'White background'. What do you think?
image

Thanks

josojo:
Already cool!

cmagan:
@alongoni thanks, looking good!
I tend to think we don't need to show the media kit image multiple times. It is enough to show the png one on the top and then explain we have also a white version available and provide download links to all version in all formats etc

Another important thing is we need to all all the added text:
on top: "Every auction that is done using Gnosis Auction can use this Media kit:"
In the bottom:
"Important notice to auction participants:

Gnosis Auction enables permissionless P2P auctions applying a batch auction mechanism without the need for intermediaries. Auctions are initiated and organised by the auctioneers. Gnosis Auction and GnosisDAO do not control, initiate or organise auctions nor are we involved with the auctioneers. Please do your own due diligence before participating in an auction.
"

henrypalacios:
@alongoni  I liked it, I'm not sure if it's just me, but I see only the first one color with shadow, is that ok?

Selection_307

alongoni:
> @alongoni I liked it, I'm not sure if it's just me, but I see only the first one color with shadow, is that ok?
> 
> Selection_307

Hey @henrypalacios, I've to added a white shadow to this color to differentiate it from the background (same color). The orange and white colors have a dark shadow that maybe is a bit soft.

henrypalacios:
> Hey @henrypalacios, I've to added a white shadow to this color to differentiate it from the background (same color). The orange and white colors have a dark shadow that maybe is a bit soft.

@alongoni And if you use a medium brightness gray, wouldn't it be more uniform for both cases?

alongoni:
Should we merge it? @cmagan

josojo (approved):
LGMT

PR #768, opened 2021-07-07
hotfix div by 0

josojo:
@elena-zh this should tackle your comment on the PR of minFunding PR #738 738

Testplan:
http://localhost:3000/auction?auctionId=17&chainId=4#topAnchor no longer breaks


elena-zh (approved):
Great!

PR #771, opened 2021-07-09
Display the current price in legend

### LINKED ISSUE #769: Display the current price in the legend of the chart - logic
Current price is added to the chart tooltip in #574.
Currently, it shows the mocked data. 
Need to display an actual value: 
 - Price + token value
- Price should be inverted if a user inverts a token pair
 
Need to discuss whether to show an estimated current price when placing an order?

henrypalacios:
closes #769

The legend of current price displayed a mock price data.

This implementation uses PR #629 as a base.

### DIFF
--- a/src/components/auction/Charts/XYChart.ts
+++ b/src/components/auction/Charts/XYChart.ts
@@ -169,6 +169,7 @@ interface DrawInformation {
   baseToken: Token
   quoteToken: Token
   chainId: ChainId
+  textAuctionCurrentPrice: string
 }
 
 const formatNumberForChartTooltip = (n: number) => {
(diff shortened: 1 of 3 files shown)


elena-zh:
Hi @henrypalacios , 
could you please create a PR-preview link for me to test these changes?

Tnx

elena-zh:
Hi @henrypalacios , I have reviewed the PR. Great job!
Trivial issues found: 
1. it would be nice to use the same rounding for the current price at the top of an auction  and in the tooltip
use the same rounding

2. It would be nice to remove brackets from the current price value as they do not present on the design
remove brackets

3. the last case is that we need to wait for the final decision about the current price displaying in the #629 task and implement changes in accordance to it. 

Thanks!

elena-zh:
Great!

Issue #672, opened 2021-06-10
Warn a user about receiving WETH/WXDAI tokens after cancellation


elena-zh:
Discussed  in  #669  https://github.com/gnosis/ido-ux/pull/669#issuecomment-858453286

When a user cancels an order in ETH, he/she receives back WETH tokens in Mainnet/Rinkeby.
When a user cancels an order in XDAI, he/she receives back WXDAI tokens in xDAI.

It would be nice to warn a user somehow about it.  As an example, add a message to the cancellation confirmation pop-up or so.

alongoni:
Currently when a user wants to cancel an order, a modal like this appears:
![GA-CancelOrder](https://user-images.githubusercontent.com/622217/123313590-0a08d780-d500-11eb-8aed-7942291cc938.png)
I think It's **too invasive** and doesn't provide too much information.

So I suggest this design that include information to review about the order that the user is cancelling. Also we add the information about receiving WETH/XDAI tokens after cancellation.
![12-B- GnosisAuction - Cancelling](https://user-images.githubusercontent.com/622217/123313639-14c36c80-d500-11eb-821c-f2bc44310ac9.png)

What do you think? @josojo @cmagan

elena-zh:
IMO, I do not think that we need the same info as in the place order confirmation modal.
I mean, It will be enough to have 2 lines with info:
 - < token > tokens sold
 - Min (max) price per < selling token >

I think there is no need in 'Min token received' field as more likely users will cancel orders in case when a price is below a current price, and they will receive nothing when an auction ends. 

Thanks

alongoni:
Thanks @elena-zh for your comments! :)
I've upgrade the design, please let me know if you think it will work better now.
![12-B- GnosisAuction - Cancelling](https://user-images.githubusercontent.com/622217/123423623-2d7e6180-d596-11eb-8c5e-41fe19f453f1.png)

elena-zh:
@alongoni , cool, thank you!
Let wait for  @josojo and @cmagan thoughts about it.

josojo:
Yes. looks really good! I think it would be nice to make somehow clear that this is a confirmation window. maybe we can call it: Confirm order Cancelation. 
Though this is quite long.

If you don't have a good suggestion, then I think we can keep it as it is

josojo:
@maria-vslvn 

This has a higher prio than the order table improvements, in case you have capacity :)

alongoni:
I think the length of the title is ok!!
[...]

PR #782, opened 2021-07-13
warn popup updated

maria-vslvn:


elena-zh:
@maria-vslvn , 
1. additional message on the pop-up should not be displayed when bidding tokens are not ETH (for Rinkeby and Mainnet networks)  and xDAI (for xDAI network)
should not be displayed

2. Token pair (DAI per GNO) should not me mocked on the pop-up and should consider < selling token > and <  buying token > values
token values

3. Label should not contain 'Min (max)' in 1 line. Min should be displayed when label is 'Min bidding price' in the price field. Max should be displayed when label is 'Max bidding price' in the price field. See my comment [here](https://github.com/gnosis/ido-ux/issues/672#issuecomment-876539487)
max

4. Warning message should not be mocked as well. It should take into consideration bidding token value:
 - `If you Cancel an order in xDAI, you will receive back WXDAI tokens in xDAI.` - for xDAI network
 - `If you Cancel an order in ETH, you will receive back WETH tokens in Rinkeby.`- for Rinkeby network
 - `If you Cancel an order in ETH, you will receive back WETH tokens in Mainnet.`- for Mainnet network

maria-vslvn:
@elena-zh please check

elena-zh:
@maria-vslvn , 
1. The 2nd warning message on the cancellation pop-up should be always displayed in all auctions: 
2nd message should present everywhere

2. Besides, neither the 1st nor the 2nd warning message are not displayed on the cancellation pop-up when  cancel orders with ETH/XDAI bidding token

3. Min 'ETH' instead of 'WETH' (and xDAI instead of WXDAI) should be displayed on the pop-up message 
no at all + ETH

4. Token's pair should be reverted as well (besides changing min/max label):
token pair min

5. Token icons should be bigger and should not be shifted towards numbers (see as an example 'Place order' pop-up)
token icons

maria-vslvn:
updated

elena-zh:
Hi @maria-vslvn , 
changes look much better, however, these issues are still not fixed: 
1. When invert price, not only token's pair should be changed, price itself should be changed as well
inverted

2. Still, i do not see an additional warning neither for ETH token in Rinkeby and Mainnet, nor for XDAI token in the XDAI
weth-eth
xdai

maria-vslvn:
@elena-zh please check

elena-zh:
@maria-vslvn , great job!

The only thing I'd like to notice, that token names on the modal are WETH and wxDAI. 

However, I'm not sure whether it is an issue.. @josojo , could you please take a look and clarify whether we should show ETH/xDAI everywhere on the modal, or leave it as it is? 
eth-weth
xdai

Also, I did not have a possibility to test changes in the Mainnet.. Alex, could you please check these changes there? 

Thanks

josojo:
Yes the behaviour would be correct with ETH/xdai instead of WETH/wxdai.

@maria-vslvn  you can just use the function: 
```
getTokenDisplay(biddingToken, chainId) or getTokenDisplay(auctioningToken, chainId)
```
to display it correctly

maria-vslvn:
@elena-zh please check

josojo (approved):
NIce. thanks. Looks good now

elena-zh:
Additional warning now is missing when cancel an order in ETH: 
image

josojo:
@maria-vslvn  could you please tackle the issue from elena?

maria-vslvn:
@elena-zh please check

elena-zh:
@maria-vslvn , still there is no additional message =(
image

Issue #765, opened 2021-07-06
Sometimes claiming transaction gets stuck 


elena-zh:
The issue is not always reproducible, but I was able to catch it at least twice

1. Place an order in an auction
2. wait till the auction ends and is settled
3. open an auction details page
4. change networks several times
5. press on the Claim button

**AR**: auction is in 'Claiming' status, however, transaction is completed, tokens are back to a user's wallet
claiming is displayed until reload the page
![claiming.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/59d5af51-a750-4814-a1e6-5f7dfad6a965)
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/694711bf-0dc0-4490-bf73-719924f17ea1)

**ER**: transaction should be ended as soon as it is finished with a success in the Etherscan

josojo:
Yes, it would be good to investigate, which actions the network switch is triggering and which ones might no longer be triggered. 
Good part is that users will probably not run into it that often, but since this wasn't any issue in the past, it might be easier to tackle the sooner we tackle it.

henrypalacios:
After some research, my first hypothesis was that it had to do with the orders ([claimableOrders](https://github.com/gnosis/ido-ux/blob/develop/src/hooks/useClaimOrderCallback.ts#L254)), associated with an account being maintained even when I change networks.
![Selection_314](https://user-images.githubusercontent.com/4270166/126023335-f6a15ada-fe0b-4683-af4f-90006fb7b512.png)

PR #786, opened 2021-07-20
Avoid write blockNumber in erroneous chainId

henrypalacios:
closes: #765

When the `chainId` is changed, the `blockNumber` is writing to the erroneous key (`chainId`) in `application` reducer state.

Selection_316


josojo (approved):
Nice, yes this could have caused many issues! thanks for hunting this one!

elena-zh:
LGTM

Issue #605, opened 2021-05-21
[auction details] Allow List Contract and Signer Address links point to mainnet all the time


tatyanakurinskaya:
1. open app
2. open some auction (ie, some auction in rinkeby) https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=100&chainId=4#topAnchor
3. click More details
4. follow link near Allow List Contract 

**Result**: https://etherscan.io/address/0x7c882f296335734b958b35da6b2595fa00043ae9
**Exp.result:** https://rinkeby.etherscan.io/address/0x7c882f296335734b958b35da6b2595fa00043ae9

**see the same for Signer Address** : https://etherscan.io/address/0xd176effb29f9a80f06db43ecfa4ef428370ed673

Moreover, these links navigate to the etherscan when an auction is opened in **xDAI network**. 
They should navigate to `https://blockscout.com/xdai/mainnet/address/.......`

ramirotw:
Removing the medium priority as this only affects the development environment

elena-zh:
@ramirotw , I would leave the med priority due to the issue is related to the xDAI chain as well. 
Links navigate to Etherscan instead of Blockscout, and that should be fixed.

PR #791, opened 2021-07-21
shows address and signer

josojo:
Draft only

### DIFF
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -431,7 +431,7 @@ const AuctionDetails = (props: Props) => {
         tooltip: 'Address of the contract managing the allow-list for participation',
         url: `https://etherscan.io/address/${auctionDetails?.allowListManager}`,
         value: `${
-          auctionDetails && auctionDetails.allowListManager == ''
+          auctionDetails && auctionDetails.allowListManager !== ''
             ? auctionDetails.allowListManager.substr(0, 6).concat('...')
             : 'None'
         }`,
@@ -441,7 +441,7 @@ const AuctionDetails = (props: Props) => {
         tooltip: 'Signer Address',
         url: `https://etherscan.io/address/${auctionDetails?.allowListSigner}`,
         value: `${
-          auctionDetails && auctionDetails.allowListSigner == ''
+          auctionDetails && auctionDetails.allowListSigner !== ''
             ? auctionDetails.allowListSigner.substr(0, 6).concat('...')
             : 'None'
         }`,


elena-zh:
Hi @josojo , 
changes look great to me. However, some minor issues: 
1. It would be nice to display addresses according to the mock up https://zpl.io/b6KvD5m
image

2. Then, it looks a bit weird to me when zeros are displayed for a public auction. Could we show None instead of zeros (like it was before)? WDYT?
image

src/components/auction/AuctionDetails/index.tsx
```diff
@@ -431,7 +431,7 @@ const AuctionDetails = (props: Props) => {
         tooltip: 'Address of the contract managing the allow-list for participation',
         url: `https://etherscan.io/address/${auctionDetails?.allowListManager}`,
         value: `${
-          auctionDetails && auctionDetails.allowListManager == ''
+          auctionDetails && auctionDetails.allowListManager !== ''
             ? auctionDetails.allowListManager.substr(0, 6).concat('...')
```
henrypalacios, line 435: Alex, what do you think if we use of the methods we already have for this? ([shorenAddress](https://github.com/gnosis/ido-ux/blob/showAddresses/src/utils/index.ts#L73) or [truncateStringInTheMiddle](https://github.com/gnosis/ido-ux/blob/showAddresses/src/utils/tools.ts#L3))
with `shortenAddress(address, 3)`  we would show 3 characters at both ends like **0x7C8...AE9**

src/components/auction/AuctionDetails/index.tsx
```diff
@@ -431,7 +431,7 @@ const AuctionDetails = (props: Props) => {
         tooltip: 'Address of the contract managing the allow-list for participation',
         url: `https://etherscan.io/address/${auctionDetails?.allowListManager}`,
```
henrypalacios, line 432: Wouldn't this be failing for other networks?

elena-zh:
@josojo , one more important case: links always navigate to the  Ehterscan.io, however, it would be great to navigate to the blockscout in xdai, and to https://rinkeby.etherscan.io/ when we open auctions there. 
https://drive.google.com/file/d/1cfyFc4xyheWcGfWkGgVkQI8DAynIkn2k/view

elena-zh:
> @elena-zh @josojo In cases where the auction is not private, it is showing `0x000...`, shouldn't it be `None`?
> Selection_319

@henrypalacios , I have mentioned about it in [here](https://github.com/gnosis/ido-ux/pull/791#issuecomment-884202180): see point 2.

elena-zh:
@josojo , great job! 
Also, regarding to this my comment https://github.com/gnosis/ido-ux/issues/791#issuecomment-884230166 and @henrypalacios [notice ](https://github.com/gnosis/ido-ux/issues/791#discussion_r674006633), I have found a separate issue for this case #605 .
So we can close the current PR and fix navigation #605 issue in another one .

Thanks

henrypalacios:
@elena-zh I had already arranged with @josojo to solve it here, could you please check #605 here.

elena-zh:
@josojo , @henrypalacios , cool!
Great job!

josojo:
great! tthanks @henrypalacios @elena-zh

PR #792, opened 2021-07-22
adding kyc link and  connect button

josojo:
This adds the kyc link that can be maintained for each auctioning token

the KYC json list should be fetched from this github repo, instead of the build, but for reviewing the PR I am pulling it via internal reference.

If the account is not yet connected, currently the interface shows that the user is not allowlisted, which is confusing. I changed that behaviour as well and show the user the connect button.

@henrypalacios any proposal for improvements are highly welcome

### DIFF
--- a/src/components/auction/OrderPlacement/index.tsx
+++ b/src/components/auction/OrderPlacement/index.tsx
@@ -49,6 +49,8 @@ const LinkCSS = css`
   color: ${({ theme }) => theme.text1};
   text-decoration: none;
   transition: color 0.05s linear;
+  font-size: 1.5em;
+  font-weight: bold;
 
   &:hover {
     color: ${({ theme }) => theme.primary2};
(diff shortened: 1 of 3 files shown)


elena-zh:
Hi @josojo , great job!
However, an empty place order section with 'Connect Wallet' button look s bit odd to me.. 
May be we should add some kind of a message 'Please connect your wallet to start' or so there?
connect a wallet

Another minor case: could we navigate a user to the https://qualify.swarm.markets/ in another tab?

Thanks!

src/components/auction/OrderPlacement/index.tsx
```diff
@@ -326,18 +341,33 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
...
     !!biddingToken.address
 
+  const auctioningTokenAddress = auctioningToken && auctioningToken?.address
+  const linkForKYC = auctioningTokenAddress ? kycLinks[auctioningTokenAddress] : null
   return (
     <>
       <Wrapper>
         {auctionInfoLoading && <InlineLoading size={SpinnerSize.small} />}
-        {!auctionInfoLoading && isPrivate && !signatureAvailable && (
+        {!auctionInfoLoading && isPrivate && !signatureAvailable && account == null && (
```
henrypalacios, line 350: it may also be confusing just to show the button `Connect Wallet`, how about changing the `account` logic to show that part?

            <PrivateWrapper>
              <LockBig />
              <TextBig>Private auction</TextBig>
              <EmptyContentTextNoMargin>
                You need to get allowed to participate.
              </EmptyContentTextNoMargin>
            </PrivateWrapper>
            {account == null ? (
              <ActionButton onClick={toggleWalletModal}>Connect Wallet</ActionButton>
            ) : (
              <EmptyContentTextSmall>
                {linkForKYC ? (
                  <ExternalLink href={linkForKYC}>Get Allowed ↗</ExternalLink>
                ) : (
                  <EmptyContentTextSmall>
                    Ask the auctioneer to get allow-listed.
                  </EmptyContentTextSmall>
                )}{' '}
              </EmptyContentTextSmall>
            )}


Selection_320

src/components/auction/OrderPlacement/index.tsx
```diff
@@ -326,18 +341,33 @@ const OrderPlacement: React.FC<OrderPlacementProps> = (props) => {
...
             <TextBig>Private auction</TextBig>
             <EmptyContentTextNoMargin>
-              You are not allowed to place an order.
+              You need to get allowed to participate.
             </EmptyContentTextNoMargin>
-            <EmptyContentTextSmall>Ask the auctioneer to get allow-listed.</EmptyContentTextSmall>
+            <EmptyContentTextSmall>
+              {linkForKYC ? (
+                <ExternalLink href={linkForKYC}>
+                  <h2>Get Allowed ↗</h2>
```
henrypalacios, line 363: Maybe the designers don't like to have a `h2` here 😂, we could get the h2 style at `LinkCSS` or maybe use the same style of `Connect Wallet button`

henrypalacios (commented):
@josojo  I can't think of another scalable solution from the front end to add KYC file (maybe later we could do it from the auction creation).

The same things that @elena-zh  already mentioned came to my mind.
1. Leave comments below
2. at least from swarm go back to the auction page maybe open a new tab

src/components/auction/OrderPlacement/index.tsx
josojo: Nice, I like this one. Leet's do it

src/components/auction/OrderPlacement/index.tsx
josojo: Yes, we should modify it. Though my css styles are very bad.
@maria-vslvn could you find a proper solution that looks somehow similar to the h2?

henrypalacios:
@elena-zh I have applied some changes, what do you think?

I have tried to keep the same pattern for the other cases

Selection_322

elena-zh:
Great job!
I have created 2 low priority enhancements related to this pr: #794, #795

PR #798, opened 2021-07-30
tooltip content fix

### LINKED ISSUE #797: Checkmark icon is not displayed in 1 line with the text on the confirmation modal
I placed/cancelled an order with long numbers in amounts; token names were long as well. As a result, text on the confirmation pop-up was shifted: 

image

**Expected**: text should be in 1 line with the icon

maria-vslvn:
Closes #797


elena-zh:
LGTM

josojo (approved):
thanks

PR #801, opened 2021-08-20
cut labels

### LINKED ISSUE #800: Auction details: issues with long token names
https://gnosis-auction.eth.link/#/auction?auctionId=37&chainId=1%23topAnchor

We need to do:

1. displaying only first 6-8 characters for the token names. -> probably easy
2. make the one bid visible in the chart. -> probably hard

Also, it would be nice to adjust the Auctions table

And a featured auction card (screenshot as an example)

maria-vslvn:
Closes #800 
tokens names were cutted up to 8th char

### DIFF
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -479,6 +479,7 @@ const AuctionDetails = (props: Props) => {
     )
   }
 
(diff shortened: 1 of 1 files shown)


josojo:
I think it's better to modify the function `getTokenDisplay` and do the cutting in there.

elena-zh:
The issue seems to be partially fixed
image
image
image

PR #802, opened 2021-08-29
Feature/800 cut lbls

### LINKED ISSUE #800: Auction details: issues with long token names

maria-vslvn:
Closes #800 
modified `gettokendisplay` method allAuctions table names cutted as well

### DIFF
--- a/src/components/auction/AuctionDetails/index.tsx
+++ b/src/components/auction/AuctionDetails/index.tsx
@@ -479,6 +479,7 @@ const AuctionDetails = (props: Props) => {
     )
   }
 
+  const getCuttedTokenDisplay = (item: string) => item.slice(0, 7)
   return (
     <Wrapper>
       <MainDetails>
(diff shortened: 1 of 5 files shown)


josojo:
image
Could we get the name in the red box also shorted?

elena-zh:
In addition to the issue reported by Alex, some issues from my side: 
1. Auction details overlap mobile widget
image

2. Not an issue, but it would be better, from my perspective, to move a token pair to another row in the Order table list header
another row

@josojo , WDYT?

josojo:
Regarding Elena's points:
1. Let's find a pragmatic solution for this, either lines breaks or shorting the number of letters of the token symbol by 1.
2. If there is an easy solution, let's do it. But not spend more than 2 hours on it.

I hope we can merge this one soon

josojo:
image
We need to also find a fix for these cards.

elena-zh:
LGTM!
@josojo , please take a look whether it looks good to you either as unfortunately I can't see this auction Featured card now.

josojo:
Thanks all. I tested it again. looks good

Issue #820, opened 2021-11-09
[Dev] Remove tooltip about supported networks in an auction details


elena-zh:
1. Open an auction details (https://ido-ux.dev.gnosisdev.com/#/auction?auctionId=71&chainId=1#topAnchor)
2. Connect a wallet in the Rinkeby
3. Hover a mouse on the tooltip near 
`Connect to the Mainnet network` message

**AR**: tooltip says that supported networks are xDAI, Mainnet and Rinkeby.
![remove tooltip.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/a0c20c85-b915-4649-a1ff-8d6efbc1b20b)

**Expected:** this tooltip is not relevant for this page as: 
1. we a currently connected to Rinkeby
2. we added a new network support: Polygon

So it would be nice to delete is at all.

Issue #822, opened 2021-11-09
Polygon: undefined is in the modal when cancel an order in another network


elena-zh:
1. Place an order for an auction in Polygon network /auction?auctionId=6&chainId=137#topAnchor
2. Switch to another network
3. Try to cancel an order

**AR**: 'In order to cancel this order, please connect to **undefined** network' in the waning message
![undefined.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/756b7bb9-f3e0-4582-89a0-ebb5a4495ede)

**ER**: 'In order to cancel this order, please connect to **Polygon** network' in the waning message

PR #826, opened 2021-11-11
Add polygon to chainNames

josojo:
fixes https://github.com/gnosis/ido-ux/issues/820 and https://github.com/gnosis/ido-ux/issues/822


elena-zh:
Hey @josojo , Polygon is added to the list of supported networks in the tooltip, but the issue with overlapping remains. I created a separate issue for this case (it is already exists on the prod) #829 

Besides, it would be nice to start Polygon with the capital letter
image
as we do for Place order notification
image

Issue #825, opened 2021-11-10
Polygon: Gas estimation error when place an order


elena-zh:
Auction details: /#/auction?auctionId=10&chainId=137#topAnchor

I'm not able to place an order for the acution die to the gas estimation error
![failed.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/0e14d1f4-ebf9-4a73-859d-4a2ebfd19503)
https://watch.screencastify.com/v/zjojaSxrsTKcx0rV9Cfs

josojo:
This needs further ingestigation, though it seems to me that this is an error from the gas estimation within metamask. To which rpc-node are you connected in metamask? https://polygon-rpc.com?

elena-zh:
@josojo , yes

josojo:
@elena-zh is this bug still happening? Or was it something that happend only during a specific time period? I have a hard time figuring out what it was

elena-zh:
@josojo , will retest soon. 
Currently, I have got stuck with pending transactions 
![image](https://user-images.githubusercontent.com/70885163/141463273-c10e9a59-2282-43dd-bf6b-b206f9d12878.png)

elena-zh:
@josojo , the issue still exists
![image](https://user-images.githubusercontent.com/70885163/141466625-45ec0dba-2185-48ec-9495-2b36d91d0efc.png)
auctionId=12&chainId=137#topAnchor

Issue #823, opened 2021-11-10
Polygon: add a possibility to pull MATIC directly to an auction


elena-zh:
When an auction buying token is WETH or XDAI, we allow to place an order with ETH and XDAI and wrap it in 1 transaction. 
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/8cfb4bda-4573-4634-9d10-c07fbd02514f)
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/05d891db-add2-45b9-8738-99599e912a94)

It would be nice to add this functionality in Polygon
Wrap unwrap buttons in the Amount field should point to quickswap?

elena-zh:
Besides, WMATIC tokens are displayed, but the app shows balances for MATIC token
![image](https://user-images.githubusercontent.com/70885163/141466459-710cf87b-f970-4906-b790-08751a21c9ec.png)

Issue #828, opened 2021-11-12
Polygon: Navigate to Polygonscan in confirmation modals


elena-zh:
View transaction details in Polygonscan when: 
 - place an order
 - cancel an order
 - claim an order

Currently, a user sees navigate to 'Etherscan'
![transaction in polygon.jpg](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/7661fcd7-3923-419d-afda-1d30f5e40311)
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/e57ec420-1a3b-4c80-b7a5-6aaa5955e573)
![image.png](https://images.zenhubusercontent.com/5f55ec970a2e890ec10f7778/40b6fa67-3cf2-4804-834e-05ff8ea42d89)

PR #830, opened 2021-11-13
Allows placing orders with matic, instead of wmatic

josojo:
I am a little bit ashamed of the dirty code, but for coding-speed, we do some compromises here

closes:  #825
closes: #823


josojo:
Merged, as customers are testing it already

elena-zh:
Btw, I have connected #828 issue to this PR as it is also fixed there.

PR #833, opened 2021-12-02
Update devguide06.md - fix typo in reserve price calculation 

cmagan:


josojo (approved):
Since it is not stated in which denomications e.g. btc/eth or eth/btc, I guess both is correct

PR #845, opened 2022-12-12
Fix issues in adding goerli support

AireshBhat:
* Fixed issue in Goerli name. Changed from "GOERLI" to "GÖRLI" wherever necessary.
* Update chain id from 4 to 5 in the connectors file.

### DIFF
--- a/src/connectors/index.ts
+++ b/src/connectors/index.ts
@@ -50,8 +50,8 @@ export const walletconnect = {
     qrcode: true,
     pollingInterval: POLLING_INTERVAL,
   }),
-  4: new WalletConnectConnector({
-    rpc: { 4: NETWORK_CONFIGS[4].rpc },
+  5: new WalletConnectConnector({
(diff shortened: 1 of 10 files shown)


src/connectors/index.ts
```diff
@@ -50,8 +50,8 @@ export const walletconnect = {
     qrcode: true,
     pollingInterval: POLLING_INTERVAL,
   }),
-  4: new WalletConnectConnector({
-    rpc: { 4: NETWORK_CONFIGS[4].rpc },
+  5: new WalletConnectConnector({
```
segun, line 53: This change will remove support for rinkeby...I think add a new option for Goerli and leave chain id 4 as it is.

AireshBhat: Right, the previous commit in this branch mentioned that Rinkeby support was to be removed in replace of Goerli. There was no reference to Rinkeby in the code apart from this file.

